PR #136, opened 2016-06-10
Update git notary

aarondl:


colstrom:
Such vendor. Much wow. Approved.

PR #137, opened 2016-06-29
HCF-678 - Acceptance tests need scp and sftp

ericpromislow:
Better to put it in the Dockerfile, but fissile currently
doesn't let us do that.  So I've added https://jira.hpcloud.net/browse/HCF-758
to request this.


viovanov:
👍

PR #141, opened 2016-07-15
HCF 703 Allow building on machines behind proxies

Mook-as:
This updates the vendored `go-dockerclient` and uses the new `args` support to set up temporary environment variables during the build for HTTP proxy support (which are not persisted into the images).  Also does the same for running the containers when running the compilation containers.


alexjh:
LGTM

PR #142, opened 2016-07-27
HCF-775 - Update run.sh to look for pre-start jobs on its own. Ignore them if in role-manifest.yml

ericpromislow:


scripts/dockerfiles/run.sh
```diff
@@ -78,9 +78,17 @@ service rsyslog start
 cron
 
 # Run custom post config role scripts
+# First run all the scripts called pre-start
+for fname in $(find /var/vcap/jobs/*/bin -name pre-start) ; do
```
Mook-as: Shouldn't we at least `{{ range $job := range .role.Jobs }}`? Just wildcarding it there seems like a potentially bad idea.

scripts/dockerfiles/run.sh
```diff
@@ -78,9 +78,17 @@ service rsyslog start
...
+    bash $fname
+done
+# And run any custom scripts other than pre-start
+
 {{ range $script := .role.PostConfigScripts}}
+{{ if not (is_pre_start $script) }}
```
Mook-as: We should fix the indenting here so it's easier to read in the template. (It'll have extra indenting in the emitted shell script, but that's okay)

scripts/dockerfiles/run.sh
```diff
@@ -78,9 +78,33 @@ service rsyslog start
...
+    echo bash {{ if not (is_abs $script) }}/opt/hcf/startup/{{ end }}{{ $script }}
     bash {{ if not (is_abs $script) }}/opt/hcf/startup/{{ end }}{{ $script }}
 {{ end }}
+{{ end }}
+
+# Run all the scripts called pre-start, but ensure consul_agent/bin/pre-start is run before others.
+# None of the other pre-start scripts appear to have any dependencies on one another.
+# Use Perl to sort consul_agent/bin/pre-start before others.
+# (Perl's sort is stable, not that it matters for `find' output.)
+function sorted-pre-start-paths()
```
Mook-as: Wouldn't it be easier to:

1. Make an empty list
2. Check if consul_agent has a pre-start script; if yes, put it into the list from step 1
3. Put everything else in the list from step 1

ericpromislow: Possibly. I gave myself two constraints: run `find' only once, and don't use a temporary file to store results.  I'll try something that doesn't introduce a dependency on Perl.

ericpromislow: Done in https://github.com/hpcloud/fissile/commit/bc6a035be84a17d5dab0c9a5d315c8b1f4868543

scripts/dockerfiles/run.sh
```diff
@@ -78,9 +78,35 @@ service rsyslog start
...
+{
+    declare -a fnames
+    idx=0
+    if [ -x /var/vcap/jobs/consul_agent/bin/pre-start ] ; then
+	fnames[$idx]=/var/vcap/jobs/consul_agent/bin/pre-start
+	idx=$((idx + 1))
+    fi
+    for fname in $(find /var/vcap/jobs/*/bin -name pre-start | grep -v '/consul_agent/bin/pre-start$') ; do
+	fnames[$idx]=$fname
```
Mook-as: You can also do `fnames[${#fnames[@]}]` but that's… a bit less readable.

Actually useful comments:
- Do we want to quote `$fname`?
- We can use `continue` instead of `grep -v`

scripts/dockerfiles/run.sh
```diff
@@ -78,9 +78,35 @@ service rsyslog start
...
+	idx=$((idx + 1))
+    done
+    echo ${fnames[*]}
```
Mook-as: There's no need to actually stash everything in a list, _then_ echo them all… we can just echo them as needed.

```
if test -x /var/vcap/jobs/consul_agent/bin/pre-start ; then
    echo /var/vcap/jobs/consul_agent/bin/pre-start 
fi
for fname in /var/vcap/jobs/*/bin/pre-start ; do
    if test "${fname}" != /var/vcap/jobs/consul_agent/bin/pre-start ; then
        echo "${fname}"
    fi
done
```

ericpromislow: In general it makes more sense to invoke 'grep' once rather than test _n_ times, but in reality there are never more than 3 pre-start files in a given role, and usually just 1, so doing individual tests are ok. Keep in mind that `test` isn't a shell builtin.

PR #143, opened 2016-07-28
Support scripts which get sourced during startup

Mook-as:
They are used to support modifying the environment variables of the processes being launched.


viovanov:
👍

PR #145, opened 2016-08-01
HCF-732 Don't sort role jobs when calculating versions

Mook-as:
The ordering of jobs within a role may be significant (for bosh-task roles where each job is run sequentially).  Package order, however, is not significant and should still be sorted.


viovanov:
👍

PR #146, opened 2016-08-30
HCF 912: Property report

andreas-kupries:


cmd/root.go
```diff
@@ -28,6 +28,7 @@ var (
 	flagConfiggin      string
 	flagLightOpinions  string
 	flagDarkOpinions   string
+	flagJSON           bool
```
viovanov: let's use an `--output` flag
It can have a default value of `human` and could also be `json` or `yaml` perhaps?

viovanov:
Can you please add a sample of the output?

cmd/root.go
andreas-kupries: Sure. Note however that currently *only* `show properties` reacts to this flag in any way, shape, or form.

andreas-kupries:
Sample of the output ... Attached to https://jira.hpcloud.net/browse/HCF-912 now.
Name is `fissile-show-properties.json`. Needs to run through `jq` to be nicely formatted.

cmd/root.go
andreas-kupries: Done with commit 46881bfff4c6d0b74fced5e03fa0dd0456d43580.
Example of yaml output attached to the ticket.

app/fissile.go
```diff
@@ -207,6 +210,78 @@ func (f *Fissile) ListJobs() error {
...
+
+			for _, property := range job.Properties {
+				f.UI.Printf("\t%s: %s\n", color.YellowString(property.Name),
+					job.Description)
+			}
+		}
+	}
+}
+
+type props map[string]map[string][]string
```
viovanov: only used twice - i don't think we need to define this for the whole package

cmd/root.go
```diff
@@ -186,6 +188,13 @@ func init() {
 		"Path to a BOSH deployment manifest file that contains properties that should not have opinionated defaults.",
 	)
 
+	RootCmd.PersistentFlags().StringP(
+		"output",
+		"o",
+		"human",
+		"Choose output format, one of human, json, or yaml (currently only 'show properties')",
```
viovanov: currently only _for_ ...

app/fissile.go
andreas-kupries: Done with commit baaf4f2cad73091c586c519c1e2c38975f90d88f

cmd/root.go
andreas-kupries: Done with commit baaf4f2cad73091c586c519c1e2c38975f90d88f

viovanov:
👍

PR #147, opened 2016-08-31
HCF-926 Force mount a tmpfs volume on /tmp when compiling

Mook-as:
When building the `tar` package, sometimes the build fails because (as far as we can tell) aufs has a bug where `dirent_t.d_ino` differs from `stat_t.st_ino`.  This causes a configure check to fail (some of the time) which ends up causing compilation failures due to not having [`tar.git#e9ddc08d`](http://git.savannah.gnu.org/cgit/tar.git/commit/?id=e9ddc08da0982f36581ae5a8c7763453ff41cfe8)

Having a tmpfs mount at `/tmp` should resolve the issue, as that would mean there is no aufs involved, and therefore no magic around `dirent` vs `stat`.

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -252,9 +252,14 @@ func (d *ImageManager) RunInContainer(containerName string, imageName string, cm
 			Privileged:     false,
 			Binds:          []string{},
 			ReadonlyRootfs: false,
+			Tmpfs:          make(map[string]string),
 		},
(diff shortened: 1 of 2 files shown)


Mook-as:
This does _not_ appear to fix the issue; going back to thinking (in the mean time, avoiding accidental merges)…

PR #148, opened 2016-08-31
HCF-908 Fix path in monit logrotate config

WGautier:
There is already a monit logrotate config created during its installation.
It looks like this:
```
/var/log/monit.log {
        rotate 4
        weekly
        minsize 1M
        missingok
        create 640 root adm
        notifempty
        compress
        delaycompress
        postrotate
                invoke-rc.d monit reload > /dev/null
        endscript
}
```
All that's needed is fixing the log path from the default to the one we set in `monitrc.erb`

### DIFF
--- a/scripts/dockerfiles/Dockerfile-base
+++ b/scripts/dockerfiles/Dockerfile-base
@@ -33,5 +33,8 @@ ADD configgin /opt/hcf/configgin/
 # Add rsyslog configuration
 ADD rsyslog_conf.tgz /
 
+# Fix monit's logrotate config (created in the `apt-get install monit` above) to our new log file location
+RUN sed -i 's/log/vcap\/monit/' /etc/logrotate.d/monit
+
 # Make logrotate run hourly, not daily
 RUN mv /etc/cron.daily/logrotate /etc/cron.hourly/logrotate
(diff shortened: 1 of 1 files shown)


viovanov:
@WGautier were you able to test that this is working?

WGautier:
@viovanov There is an issue with the ubuntu docker image having a policy.rc that prevents monit from reloading. I'm looking into it.

WGautier:
Ok so, the `invoke-rc.d monit reload` was being blocked by policy-rc.d. Without a reload, monit kept writing to the old rotated log file. Bad.

I found the reason why the default ubuntu docker image has this blocking policy-rc.d in a comment of the policy-rc.d of the debian docker image:

```
# For most Docker users, "apt-get install" only happens during "docker build",
# where starting services doesn't work and often fails in humorous ways. This
# prevents those failures by stopping the services from attempting to start.
```

I imagine this is also the reason it's there in ubuntu's image. It should be fine for us to re-allow services restarting after having installed all packages via `apt-get install`.

scripts/dockerfiles/Dockerfile-base
```diff
@@ -33,5 +33,11 @@ ADD configgin /opt/hcf/configgin/
 # Add rsyslog configuration
 ADD rsyslog_conf.tgz /
 
+# Fix monit's logrotate config (created in the `apt-get install monit` above) to our new log file location
+RUN sed -i 's/log/vcap\/monit/' /etc/logrotate.d/monit
+
+# Allow monit to reload during logrotate
```
viovanov: Can you please add more detail to this comment.
The info you have in the conversation is very useful.
Please explain why the policy is the way it is, why we're changing it and link back to the online resources you've found.

viovanov:
👍

PR #150, opened 2016-08-31
HCF-926 tar fails to compile

Mook-as:
Let's try a volume at `/var/vcap/sources` ­— that _should_ be where configure is running. Probably.

Not for merge yet (this is to cause CI to happen)

### DIFF
--- a/compilator/compilator.go
+++ b/compilator/compilator.go
@@ -473,12 +473,14 @@ func (c *Compilator) compilePackage(pkg *model.Package) (err error) {
 	mounts := map[string]string{
 		c.getTargetPackageSourcesDir(pkg): docker.ContainerInPath,
 		c.getPackageCompiledTempDir(pkg):  docker.ContainerOutPath,
+		"source_mount":                    "/var/vcap/source/",
 	}
 	exitCode, container, err := c.DockerManager.RunInContainer(docker.RunInContainerOpts{
 		ContainerName: containerName,
 		ImageName:     c.BaseImageName(),
 		Cmd:           []string{"bash", containerScriptPath, pkg.Name, pkg.Version},
 		Mounts:        mounts,
+		Volumes:       map[string]map[string]string{"source_mount": nil},
 		KeepContainer: c.keepContainer,
 		StdoutWriter:  stdoutWriter,
 		StderrWriter:  stderrWriter,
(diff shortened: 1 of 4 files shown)


compilator/compilator.go
```diff
@@ -471,16 +470,21 @@ func (c *Compilator) compilePackage(pkg *model.Package) (err error) {
...
-		c.getTargetPackageSourcesDir(pkg),
-		c.getPackageCompiledTempDir(pkg),
-		c.keepContainer,
-		stdoutWriter,
-		stderrWriter,
-	)
+	mounts := map[string]string{
+		c.getTargetPackageSourcesDir(pkg): docker.ContainerInPath,
+		c.getPackageCompiledTempDir(pkg):  docker.ContainerOutPath,
+		"source_mount":                    "/var/vcap/source/",
```
viovanov: where does this live?
shouldn't we have a `getVCAPSourceTempDir` that returns an absolute path?
also a constant for `/var/vcap/source/`

Mook-as: It's a temporary docker volume, which lives wherever docker wants it to live.  Usually, it's in `/var/lib/docker/volumes/` somewhere. `source_mount` is just a name to figure out volumes.  (I actually munge it in the callee to include the container name… also happy to stick in a UUID or something to avoid reuse.)

Yeah, we should add a constant for that.  Was just hurrying up the thing to make sure it actually fixes the problems, first.

viovanov: how does docker know it's a "temporary" volume?
Because it's not an absolute path?
If so, can you please comment the code with that info?
Also, UUID sounds good.

compilator/compilator_test.go
```diff
@@ -182,108 +182,134 @@ func TestCompilationRoleManifest(t *testing.T) {
 	}
 }
 
-func getContainerIDs(testRepository string) (string, error) {
-	val, err := exec.Command("docker", "ps", "-a", "--format={{.ID}}/{{.Image}}").CombinedOutput()
```
viovanov: THANK YOU!

viovanov: oh, i see this is a test - that line freaked me out more than it should have :)

compilator/compilator.go
Mook-as: As far as docker is concerned, it's not temporary.  I just happen to make sure I delete it afterwards.  (I'll add appropriate comments on this.)

Will add some randomness to the volume name.

compilator/compilator_test.go
Mook-as: Yes, it's a test.  I still couldn't stand having it around, though, because, really, we have this whole library thing…

compilator/compilator.go
Mook-as: Sample volume name, after changes: `volume_fissile-cbase-fissile-2.0.1-14.g14c8518.HCF-926-tar-fails-to-compile-cf-239-dev.1-pkg-ruby-2.2.5_source_mount-8d9a1db0-45cb-4625-82d6-9d2914e9de19`
(Includes: `volume_` prefix; container name including `cbase` image with fissile version, package name; `source_mount` string; randomly generated UUID.)

compilator/compilator_test.go
viovanov: 👍

viovanov:
👍

PR #151, opened 2016-09-13
HCF-958 monit graceful exit

Mook-as:
* Changes `monit` to be pid 1, instead of being a child of `run.sh`
 * This means it can handle signals from docker correctly (e.g. `docker rm` will stop it quickly, as it receives `SIGTERM`)
* This also means moving `cron` and `rsyslog` to be managed by `monit`, so that they don't die when their parent isn't the shell
 * I _think_ it has to do with the tty, but I'm not sure; I just got zombied rsyslog a lot.
* Which in turn means extracting `rsyslog_conf.tgz` to make it actually possible to figure out what I'm doing


viovanov:
👍

PR #155, opened 2016-09-23
HCF-988 `fissile show properties --output-json` failing with hash defaults

Mook-as:
We were failing to display properties which had hashes as default values when JSON output was requested, as YAML deserializes into `map[interface{}]interface{}` but JSON requires `map[string]interface{}` to marshal the data.

This adds a helper method that converts deserialized YAML data into something that JSON can marshal.  This unfortunately involves runtime reflection.

Also updates github.com/strtchr/testify to get `assert.JSONEq`


util/json_helper.go
```diff
@@ -0,0 +1,69 @@
...
+	return fmt.Sprintf("Error marshalling JSON: Failed to convert keys in path %s: %s",
+		strings.Join(e.keys, "."), e.msg)
+}
+
+// jsonMarshalHelper converts a map from having interface{} keys to string keys
+func jsonMarshalHelper(input interface{}) (interface{}, *jsonMarshalError) {
+	value := reflect.ValueOf(input)
+	switch value.Kind() {
+	case reflect.Map:
+		if value.Kind() != reflect.Map {
```
andreas-kupries: Is this even possible given that we are in the `reflect.Map` branch of a switch testing `value.Kind()` ?

andreas-kupries:
LGTM now.

PR #156, opened 2016-09-28
[HCF-900] Added command 'show descriptions'.

andreas-kupries:
Analogous to 'show properties', it returns descriptions instead of defaults.

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -220,10 +220,6 @@ func (f *Fissile) ListProperties(outputFormat string) error {
 	case "human":
 		f.listPropertiesForHuman()
 	case "json":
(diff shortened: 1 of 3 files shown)


andreas-kupries:
Using lambda, i.e. unnamed function it is possible to factor the general traversal of releases, jobs, and properties into a single generic function. Defered.

andreas-kupries:
I am putting this one down as `stalled` now, as well as the https://github.com/hpcloud/hcf/pull/670. With https://github.com/hpcloud/hcf/pull/671 pretty much filling all the descriptions now this becomes irrelevant. Tagged however to keep around just in case.

PR #157, opened 2016-10-07
HCF-949 build role images in parallel

Mook-as:
Turns out we already have a thing that builds images in parallel, it just does non-release images and we never use it.

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -162,9 +162,13 @@ func (d *ImageManager) FindImage(imageName string) (*dockerclient.Image, error) 
 }
 
 // HasImage determines if the given image already exists in Docker
-func (d *ImageManager) HasImage(imageName string) bool {
-	_, err := d.FindImage(imageName)
-	return (err == nil)
+func (d *ImageManager) HasImage(imageName string) (bool, error) {
+	if _, err := d.FindImage(imageName); err == ErrImageNotFound {
+		return false, nil
+	} else if err != nil {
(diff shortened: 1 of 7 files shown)


docker/docker.go
```diff
@@ -161,6 +161,12 @@ func (d *ImageManager) FindImage(imageName string) (*dockerclient.Image, error)
 	return image, nil
 }
 
+// HasImage determines if the given image already exists in Docker
+func (d *ImageManager) HasImage(imageName string) bool {
+	_, err := d.FindImage(imageName)
+	return (err == nil)
```
viovanov: This is not 100% correct - see `FindImage` above.
Docker may be replying with other errors.
This helper is not proper. The callers will be ignoring errors.
We should remove this.

Mook-as: Would having it return `(bool, error)` and specifically checking for `docker.ErrImageNotFound` be okay?

(This was added so we can shim `FindImage()` without having to pull in the `docker` package.)

Mook-as: Addressed in 93e35bbfdeafbff5f32116e3b0982885593d9f8f

Also, it turns out that the [current code](https://github.com/hpcloud/fissile/blob/afffe8af344c4fc33557b722df5dfa8177b7bdb4/app/fissile.go#L371) also throws away the error (but at least it assumes we _will_ need to rebuild).

viovanov: 👍

PR #159, opened 2016-10-14
HCF-583 Put all packages into one layer

Mook-as:
This combines all the packages (and specs) into one layer, which is then shared by the various roles.  It reduces the _total_ disk usage of all the images in aggregate as shared packages (e.g. ruby) only take up disk space once, rather than being stored separately per role.  This does mean though that each image separately would be larger (as they now contain every package for the whole thing).

Estimate is something like 40G total -> 8G in `/var/lib/docker`.


Mook-as:
Made a [Jenkins run](https://jenkins.issueses.io/view/HCF/job/hcf-vagrant-in-cloud-develop/301/console) and am now reasonably happy with the changes.  Will probably do a follow-up to reduce copying in the package image creation, though. (But that can be separate from this PR — it should already be better now.)

builder/role_image.go
```diff
@@ -146,6 +223,7 @@ func (r *RoleImageBuilder) CreateDockerfileDir(role *model.Role, jsonSpecsDir st
 	if err := os.MkdirAll(jobsDir, 0755); err != nil {
 		return "", err
 	}
+	jsonSpecsDir := filepath.Join(rootDir, "opt/hcf/specs")
```
viovanov: What is the specs dir?

builder/role_image.go
```diff
@@ -315,21 +406,21 @@ func (j roleBuildJob) Run() {
 			j.resultsCh <- err
 			return
 		} else if hasImage {
-			j.ui.Println("Skipping image build because it exists")
+			j.ui.Printf("Skipping build of role images %s because it exists\n", color.YellowString(j.role.Name))
```
viovanov: `images` should be singular, correct?

builder/role_image.go
```diff
@@ -50,25 +50,120 @@ type RoleImageBuilder struct {
...
+}
+
+// CreatePackagesDockerBuildDir generates a Dockerfile and assets for the shared packages layer and returns a path to the dir
+func (r *RoleImageBuilder) CreatePackagesDockerBuildDir(roleManifest *model.RoleManifest, lightManifestPath, darkManifestPath string) (string, error) {
+	if len(roleManifest.Roles) == 0 {
+		return "", fmt.Errorf("No roles to build")
+	}
+
+	succeeded := false
+	buildDir, err := ioutil.TempDir(r.targetPath, "role-packages")
```
Mook-as: I'm planning on having a follow-up PR to put it somewhere in the compilation cache's ancestry (so that we reduce copying).

Using a temp dir everywhere was to guard against things going really wrong and accidentally reusing cruft from a previous run; I'd rather we leave things behind on accident (though we shouldn't) than making bad builds.

builder/role_image.go
```diff
@@ -50,25 +50,120 @@ type RoleImageBuilder struct {
...
+	if err := configStore.WriteBaseConfig(roleManifest); err != nil {
+		return "", fmt.Errorf("Error writing base config: %s", err.Error())
+	}
+
+	// Copy packages
+	r.ui.Println("Copying compiled packages...")
+	copiedFingerprints := make(map[string]bool)
+	for _, role := range roleManifest.Roles {
+		for _, job := range role.Jobs {
+			for _, pkg := range job.Packages {
```
Mook-as: This will no longer exist after the next PR, at which point it stops mattering :grinning: 

Also, I don't think doing it in parallel would have been useful: since it's all disk-bound, doing it in parallel just means more seeks (on spinning rust) instead of building things faster.  Parallelization is way more useful for CPU-bound tasks.

builder/role_image.go
```diff
@@ -50,25 +50,120 @@ type RoleImageBuilder struct {
...
 
 // CreateDockerfileDir generates a Dockerfile and assets in the targetDir and returns a path to the dir
-func (r *RoleImageBuilder) CreateDockerfileDir(role *model.Role, jsonSpecsDir string) (string, error) {
+func (r *RoleImageBuilder) CreateDockerfileDir(role *model.Role, baseImageName string) (string, error) {
 	if len(role.Jobs) == 0 {
 		return "", fmt.Errorf("Error - role %s has 0 jobs", role.Name)
 	}
 
+	succeeded := false
+	roleDir, err := ioutil.TempDir(r.targetPath, fmt.Sprintf("role-%s", role.Name))
```
Mook-as: (See [above](https://github.com/hpcloud/fissile/pull/159#discussion_r83737996))

builder/role_image.go
Mook-as: We were, yes.

builder/role_image.go
Mook-as: Err, sorry, you're right; per job.

I can move it out if desired, but we generate them all in one go anyways, so doing it per-role-image seemed pointless.

PR #161, opened 2016-10-19
HCF-583 Reduce copying when building packages layer image

Mook-as:
Instead of copying all the compiled files to a scratch workspace and telling docker to build from there, just build the tar file ourselves (which is what would have happened before too).  This reduces a set of read/write that didn't need to happen.

In tests, it reduced rebuilding the packages layer image from ~4:30 to ~3:40.

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -341,35 +341,35 @@ func (f *Fissile) GeneratePackagesRoleImage(repository string, roleManifest *mod
 	}
 
 	f.UI.Printf("Creating Dockerfile for packages layer...\n")
-	buildDir, err := packagesImageBuilder.CreatePackagesDockerBuildDir(roleManifest, lightManifestPath, darkManifestPath)
+	tarStream, errors, err := packagesImageBuilder.CreatePackagesDockerStream(roleManifest, lightManifestPath, darkManifestPath)
 	if err != nil {
 		return err
 	}
-	defer os.RemoveAll(buildDir)
+	defer tarStream.Close()
 
 	if noBuild {
 		f.UI.Println("Skipping packages layer docker image build because of --no-build flag.")
 		return nil
 	}
 
-	if !strings.HasSuffix(buildDir, string(os.PathSeparator)) {
-		buildDir = fmt.Sprintf("%s%c", buildDir, os.PathSeparator)
-	}
-	f.UI.Printf("Building packages layer docker image %s in %s ...\n",
-		color.YellowString(packagesLayerImageName),
-		color.YellowString(buildDir))
+	f.UI.Printf("Building packages layer docker image %s ...\n",
+		color.YellowString(packagesLayerImageName))
 	log := new(bytes.Buffer)
 	stdoutWriter := docker.NewFormattingWriter(
 		log,
 		docker.ColoredBuildStringFunc(packagesLayerImageName),
 	)
 
-	err = dockerManager.BuildImage(buildDir, packagesLayerImageName, stdoutWriter)
+	err = dockerManager.BuildImageFromStream(tarStream, packagesLayerImageName, stdoutWriter)
 	if err != nil {
 		log.WriteTo(f.UI)
 		return fmt.Errorf("Error building packages layer docker image: %s", err.Error())
 	}
 	f.UI.Println(color.GreenString("Done."))
+
+	for err = range errors {
+		return err
+	}
 
 	return nil
 }
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -367,11 +367,7 @@ func (f *Fissile) GeneratePackagesRoleImage(repository string, roleManifest *mod
 	}
 	f.UI.Println(color.GreenString("Done."))
 
-	for err = range errors {
-		return err
-	}
-
-	return nil
+	return <-errors
 }
 
 // GenerateRoleImages generates all role images using dev releases
--- a/builder/packages_image.go
+++ b/builder/packages_image.go
@@ -51,10 +51,12 @@ func (w *tarWalker) walk(path string, info os.FileInfo, err error) error {
 	if err != nil {
 		return err
 	}
+
 	header, err := tar.FileInfoHeader(info, "")
 	if err != nil {
 		return err
 	}
+
 	if (info.Mode() & os.ModeSymlink) != 0 {
 		linkname, err := os.Readlink(path)
 		if err != nil {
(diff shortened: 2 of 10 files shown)


builder/packages_image.go
```diff
@@ -39,86 +40,135 @@ func NewPackagesImageBuilder(repository, compiledPackagesPath, targetPath, fissi
...
+		}
+		walker := &tarWalker{stream: tarStream, root: specDir, prefix: "specs"}
+		if err := filepath.Walk(specDir, walker.walk); err != nil {
+			errors <- err
+			return
+		}
 
-				copiedFingerprints[pkg.Fingerprint] = true
+		// Collect compiled packages
+		foundFingerprints := make(map[string]bool)
```
aarondl: Typically when using a map as a set people will use an empty struct for a zero allocation type of thing: `map[string]struct{}`. This can generally be used to replace most uses of `map[string]bool`. It's not strictly necessary but just letting you know.

builder/packages_image_test.go
```diff
@@ -1,18 +1,62 @@
...
 	"github.com/hpcloud/termui"
 	"github.com/stretchr/testify/assert"
 )
 
-func TestCreatePackagesDockerBuildDir(t *testing.T) {
+func TestGenerateDockerfile(t *testing.T) {
+	assert := assert.New(t)
+
+	ui := termui.New(
+		os.Stdin,
```
aarondl: Do you really want to hook up stdin from the testing process to this?

builder/packages_image_test.go
```diff
@@ -1,18 +1,62 @@
...
+
+	compiledPackagesDir := filepath.Join(workDir, "../test-assets/tor-boshrelease-fake-compiled")
+	targetPath, err := ioutil.TempDir("", "fissile-test")
+	assert.NoError(err)
+	defer os.RemoveAll(targetPath)
+
+	packagesImageBuilder, err := NewPackagesImageBuilder("foo", compiledPackagesDir, targetPath, "3.14.15", ui)
+	assert.NoError(err)
+
+	baseImage := "scratch:latest"
```
aarondl: Kind of a useless variable. You could argue it's easy to see and easy to change, but you've lost that argument by using the hardcoded string again in the `assert.Equal` below. Use it in both places, or move it into `generateDockerfile()` IMHO

builder/packages_image_test.go
```diff
@@ -39,40 +83,82 @@ func TestCreatePackagesDockerBuildDir(t *testing.T) {
...
+	tarReader := tar.NewReader(tarStream)
+	for {
+		header, err := tarReader.Next()
+		if err == io.EOF {
+			break
+		}
+		if !assert.NoError(err) {
+			break
+		}
+		if expected, ok := expectedContents[header.Name]; ok {
```
aarondl: You could also invert this if statement to save some identation.

builder/packages_image_test.go
```diff
@@ -39,40 +83,82 @@ func TestCreatePackagesDockerBuildDir(t *testing.T) {
...
+			actual, err := ioutil.ReadAll(tarReader)
+			assert.NoError(err)
+			if strings.HasSuffix(header.Name, ".json") {
+				assert.JSONEq(expected, string(actual))
+			} else {
+				var expectedLines, actualLines []string
+				for _, line := range strings.Split(expected, "\n") {
```
aarondl: 3 occurrences of this code. Can we make it a function?

docker/docker_test.go
```diff
@@ -435,3 +438,50 @@ func TestFormatWriterOneLine(t *testing.T) {
...
+			Name:     "Dockerfile",
+			Mode:     0644,
+			Size:     int64(contents.Len()),
+			Typeflag: tar.TypeReg,
+		}
+		assert.NoError(tarStream.WriteHeader(&header))
+		_, err := io.Copy(tarStream, contents)
+		assert.NoError(err)
+		assert.NoError(tarStream.Close(), "Error closing tar stream")
+		assert.NoError(pipeWriter.Close(), "Error closing tar pipe")
```
aarondl: Reader close?

app/fissile.go
```diff
@@ -341,36 +341,36 @@ func (f *Fissile) GeneratePackagesRoleImage(repository string, roleManifest *mod
...
-	err = dockerManager.BuildImage(buildDir, packagesLayerImageName, stdoutWriter)
+	err = dockerManager.BuildImageFromStream(tarStream, packagesLayerImageName, stdoutWriter)
 	if err != nil {
 		log.WriteTo(f.UI)
 		return fmt.Errorf("Error building packages layer docker image: %s", err.Error())
 	}
 	f.UI.Println(color.GreenString("Done."))
 
+	for err = range errors {
+		return err
```
Mook-as: Oh, I totally forgot that receiving from a closed channel returns successfully.

builder/packages_image.go
```diff
@@ -39,86 +40,135 @@ func NewPackagesImageBuilder(repository, compiledPackagesPath, targetPath, fissi
...
+	pipeReader, pipeWriter := io.Pipe()
+	errors := make(chan error)
 
-				compiledDir := pkg.GetPackageCompiledDir(p.compiledPackagesPath)
-				if err := util.ValidatePath(compiledDir, true, fmt.Sprintf("compiled dir for package %s", pkg.Name)); err != nil {
-					return "", err
-				}
+	go func() {
+		defer close(errors)
+		defer pipeWriter.Close()
```
Mook-as: No, because that gets passed to the docker client. They can close it when they're done reading (when it hits EOF).  In fact, I _shouldn't_ close it for them, otherwise they'd get a short read.

builder/packages_image.go
```diff
@@ -39,86 +40,135 @@ func NewPackagesImageBuilder(repository, compiledPackagesPath, targetPath, fissi
...
-						IgnoreDanglingSymlinks: false,
-					},
-				)
-				if err != nil {
-					return "", err
-				}
+		// Generate configuration
+		specDir, err := ioutil.TempDir("", "fissile-spec-dir")
+		if err != nil {
+			errors <- err
```
Mook-as: That _is_ tempting; I did hate the `errors <- foo ; return` I had.

docker/docker.go
```diff
@@ -148,6 +148,40 @@ func (d *ImageManager) BuildImage(dockerfileDirPath, name string, stdoutWriter i
...
+func (d *ImageManager) BuildImageFromStream(stream io.Reader, name string, stdoutWriter io.Writer) error {
+	bio := dockerclient.BuildImageOptions{
+		Name:         name,
+		NoCache:      true,
+		InputStream:  stream,
+		OutputStream: stdoutWriter,
+	}
+
+	for _, envVar := range []string{"http_proxy", "https_proxy", "no_proxy"} {
+		for _, name := range []string{strings.ToLower(envVar), strings.ToUpper(envVar)} {
```
Mook-as: Yeah, I just copied this code from the other function.
I wonder if we should make that call this…

aarondl:
LGTM

PR #163, opened 2016-10-20
HCF-583 Attempt to re-use layers for the package layer if possible

Mook-as:
We now try to find existing images derived from the base layer that has fingerprints we desire, and build on top of those.

Pulls in gomock / mockgen to test `github.com/hpcloud/fissile/docker` with


viovanov:
👍

PR #164, opened 2016-10-21
HCF-1056 disable monit logging to disk

Mook-as:
It's already being logged to stdout (which gets handled by our docker logging infrastructure), there's no need to keep a copy on disk in a container.


viovanov:
👍

PR #168, opened 2016-10-28
builder: package images test: fix flaky test

Mook-as:
We don't care about the _order_ of the fingerprints when used as docker image labels, so sort them before comparison.


ericpromislow:
LGTM

PR #169, opened 2016-11-01
HCF-1055 - Add more pieces to a role's signature to determine when an image needs rebuilding.

ericpromislow:
This doesn't change usage of `fissile build images` as it only uses an existing global flag,
but it does change the interface to `fissile.GenerateRoleImages`

The gist: when running `fissile build images`, if a role exists but fissile detects that a crucial dependency has changed, such as a change in a property set, property value, the job set, or the job's scripts, it will rebuild the image.


aarondl:
Talked to Eric about possibly using the sha that's coming out of the tag for the image instead of a configuration directory + flag. He said he'd look into it.

andreas-kupries:
Note: `make format` complains about indentation in `docker_test.go`.
This seems to be a pre-existing issue, as the commit did not change that file.

test-assets/detect-changes-configs/ntp-release/jobs/ntpd/spec~
```diff
@@ -0,0 +1,15 @@
+---
```
andreas-kupries: Why do we have emacs backup files in the commit ?
Four of them in total
```
./test-assets/detect-changes-configs/ntp-release/jobs/ntpd/spec~
./test-assets/detect-changes-configs/small01/after-dark-opinion-value-change/dark-opinions.yml~
./test-assets/detect-changes-configs/small01/after-dark-opinion-drop/dark-opinions.yml~
./test-assets/detect-changes-configs/small01/after-dark-opinion-add/dark-opinions.yml~
```

andreas-kupries:
I am not sure how the checks pass.
I have the unit tests failing
```
--- FAIL: TestCreatePackagesDockerStream (0.01s)
        Error Trace:    packages_image_test.go:184
	Error:		Should be empty, but was map[packages-src/b9973278a447dfb5e8e67661deaa5fe7001ad742/bar:0x495850 Dockerfile:0x494f90]
	Messages:	Missing files in tar stream
		
        Error Trace:    packages_image_test.go:186
	Error:		Received unexpected error "API error (500): Invalid filter 'since'\n"
		
FAIL
```

andreas-kupries:
I am on the proper commit, i.e. 9263e55077ab62c03d8f58a5a691968ebb722562

test-assets/detect-changes-configs/ntp-release/jobs/ntpd/spec~
ericpromislow: Thanks.  I had been watching for them, and thought I had them all. Didn't remember to search in the 75+ file commit.

ericpromislow:
`make format` runs without any output on my box.

`make test` passes on my box.

I'll need to dig in to see the problem.

model/roles.go
```diff
@@ -168,26 +311,80 @@ func (r *Role) GetScriptPaths() map[string]string {
...
+	writeScripts(hasher, scriptParentDir, r.PostConfigScripts, "PostConfigScripts")
+	io.WriteString(hasher, "<jobs>")
 	// Jobs are *not* sorted because they are an array and the order may be
 	// significant, in particular for bosh-task roles.
 	for _, job := range r.Jobs {
-		roleSignature = fmt.Sprintf("%s\n%s", roleSignature, job.SHA1)
+		io.WriteString(hasher, job.SHA1)
 		packages = append(packages, job.Packages...)
+		hasher.Write([]byte("<properties>"))
+		for _, property := range job.Properties {
```
andreas-kupries: Within a single job, does the order of properties matter ?
If not we should sort them here, I think.

andreas-kupries:
Could it be something in my environment which confuses the tests ?
(I am on go1.6)

andreas-kupries:
Ok, it was the fact of me using a very old `goimports` (Jan 6, 2016). Upgraded that tool and the formatting complaint went away.

test-assets/detect-changes-configs/ntp-release/jobs/ntpd/spec~
andreas-kupries: I have two aliases for convenience:
The first simply prints all under CWD. The second removes them.
 ```
alias ftilde=`find    . -name "*~" -print'
alias rtilde=`find    . -name "*~" -exec rm "{}" \;'
```

File: builder/role_image.go [hunk 3]
```diff
@@ -24,8 +24,9 @@ import (
 )
 
 const (
-	binPrefix             = "bin"
-	jobConfigSpecFilename = "config_spec"
+	binPrefix               = "bin"
+	jobConfigSpecFilename   = "config_spec"
+	roleCacheConfigBasename = "roleConfigs.json"
```
andreas-kupries: This new const seems to be remnant of when the change used a separate cache file for hashes.
AFAICT it is not needed in the current code.

builder/role_image.go
```diff
@@ -49,6 +50,9 @@ type RoleImageBuilder struct {
 	ui                   *termui.UI
 }
 
+// HashCacheType is used to map role names to config fingerprints
+type HashCacheType map[string]string
```
andreas-kupries: I see no uses of this type in the code either.

builder/role_image.go
```diff
@@ -309,12 +313,18 @@ func (j roleBuildJob) Run() {
...
+		j.resultsCh <- err
+		return
+	}
+	roleImageName := GetRoleDevImageName(j.repository, j.role, versionHash)
 	if !j.force {
 		if hasImage, err := j.dockerManager.HasImage(roleImageName); err != nil {
 			j.resultsCh <- err
 			return
 		} else if hasImage {
+
```
andreas-kupries: Style nit, I would consider this empty line as superfluous.

make/test
```diff
@@ -4,3 +4,5 @@ set -o errexit
 
 go generate ./docker/ # for mocks
 go test -cover $(go list -f '{{ .ImportPath }}' ./... | sed '/fissile[/]scripts/d ; /\/vendor\//d')
+#go test -v -run TestNoImageChangesAfterDarkOpinionValueChanged -cover $(go list -f '{{ .ImportPath }}' ./... | sed '/fissile[/]scripts/d ; /\/vendor\//d')
+#go test -v -run TestGetRoleManifestDevPackageVersion  -cover $(go list -f '{{ .ImportPath }}' ./... | sed '/fissile[/]scripts/d ; /\/vendor\//d')
```
andreas-kupries: This seems to be debug code which got into the commit.

model/roles.go
```diff
@@ -134,22 +145,154 @@ func LoadRoleManifest(manifestFilePath string, releases []*Release) (*RoleManife
...
+		return err
+	}
+	manifestConfig.lightOpinions = opinions
+	opinions, err = manifestConfig.loadOpinionsFromPath(darkManifestPath)
+	if err != nil {
+		return err
+	}
+	manifestConfig.darkOpinions = opinions
+	manifestConfig.globalDefaultValues = getDefaultValuesHash(m.Configuration.Templates)
+	m.globalConfig = &manifestConfig
```
andreas-kupries: I believe it would be nicer to write the above as
```
m.globalConfig = &manifestConfiguration{
lightOpinions: lightOpinions,
darkOpinions: darkOpinions,
}
```
using two helper variables for the loaded data, and defers the allocation of the structure to after everything is loaded without error.

model/roles.go
```diff
@@ -134,22 +145,154 @@ func LoadRoleManifest(manifestFilePath string, releases []*Release) (*RoleManife
...
 	sort.Sort(roles)
 
 	hasher := sha1.New()
-	hasher.Write([]byte(extra))
+	io.WriteString(hasher, extra)
 
 	for _, role := range roles {
-		hasher.Write([]byte(role.GetRoleDevVersion()))
+		s, err := role.GetRoleDevVersion()
+		if err == nil {
```
andreas-kupries: When an error happens it seems to be ignored here, now.
Is that ok ? All other places using `GetRoleDevVersion()` I can find report its errors.

builder/role_image.go
ericpromislow: 5bd6d68

builder/role_image.go
ericpromislow: 33ba652

builder/role_image.go
ericpromislow: Agreed, it should have been removed

8fb1dc0

model/roles.go
```diff
@@ -168,26 +311,80 @@ func (r *Role) GetScriptPaths() map[string]string {
...
+				io.WriteString(hasher, "</dark>")
+			}
+			val, err := lookupProperty(property.Name, lightOpinions, localDefaultValues, globalDefaultValues)
+			if err != nil {
+				continue
+			}
+			io.WriteString(hasher, property.Name)
+			io.WriteString(hasher, "=")
+			io.WriteString(hasher, val)
+			io.WriteString(hasher, "</properties>")
```
viovanov: xml?

model/roles.go
ericpromislow: I don't see this in my code.  I see this:```
	for _, role := range roles {
		s, err := role.GetRoleDevVersion()
		if err == nil {
			io.WriteString(hasher, s)
		}
	}
```

andreas-kupries: Yes, that is the code I see.
What happens for `err != nil` here ?
Nothing. It ignores it.
Should we do that ?

ericpromislow: This code is obsolete.  GetRoleDevVersion will be rewritten to just return a string.

aarondl:
This PR is being closed in favor of HCF-1124

PR #171, opened 2016-11-08
[HCF-1080] New sub command "build cleancache". 

andreas-kupries:
Removes package data not referenced (anymore) by the loaded release(s) from the local compilation cache.

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


app/fissile.go
```diff
@@ -304,6 +305,65 @@ func (f *Fissile) Compile(repository, targetPath, roleManifestPath string, worke
...
+	if len(f.releases) == 0 {
+		return fmt.Errorf("Releases not loaded")
+	}
+
+	referenced := make(map[string]int)
+	for _, release := range f.releases {
+		for _, pkg := range release.Packages {
+			referenced[pkg.Version] = 1
+		}
+
```
jandubois: Spurious empty line?

cmd/build-cleancache.go
```diff
@@ -0,0 +1,31 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// buildCleancacheCmd represents the cleancache command
```
jandubois: Shouldn't `cache` be uppercase? `buildCleanCacheCmd`?

cmd/build-cleancache.go
```diff
@@ -0,0 +1,31 @@
...
+This command will inspect the compilation cache populated by its sibling "packages"
+and remove all which are not required anymore.`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+
+		if err := fissile.LoadReleases(
+			flagRelease,
+			flagReleaseName,
+			flagReleaseVersion,
+			flagCacheDir,
+		); err != nil {
```
jandubois: I think putting the assignment on its own makes it easier to read:

```
err := fissile.LoadReleases(...)
if err != nil {
}
```

app/fissile.go
```diff
@@ -304,6 +305,65 @@ func (f *Fissile) Compile(repository, targetPath, roleManifestPath string, worke
...
+		f.UI.Printf("- Removing %s\n", color.YellowString(key))
+		if err := os.RemoveAll(cache); err != nil {
+			return err
+		}
+		removed++
+	}
+
+	if removed == 0 {
+		f.UI.Printf("Nothing found to remove\n")
```
jandubois: Use `Print` or even `Println` if there is no format string.

app/fissile.go
andreas-kupries: Yes. Fixed local.

app/fissile.go
andreas-kupries: Will do

cmd/build-cleancache.go
andreas-kupries: Will do.

cmd/build-cleancache.go
andreas-kupries: This is copied from all the other `build-*` commands. Loath to make this inconsistent.

app/fissile.go
andreas-kupries: Addressed with commit 86daf7d77259018bb3bfb5eab56ce201e4c22d2b

cmd/build-cleancache.go
andreas-kupries: Addressed with commit 86daf7d77259018bb3bfb5eab56ce201e4c22d2b

cmd/build-cleancache.go
jandubois: @aarondl agrees that there is no reason to do this, so I think it should be changed:

https://helion-stackato.slack.com/archives/hcf/p1478652709018844

Maybe add a separate commit that fixes this everywhere?

jandubois:
I've now tested the code as well, and it seems to work as expected. So I'm fine merging it when the last style issue has been fixed.

cmd/build-cleancache.go
andreas-kupries: Yep. Already agreed on Slack, now here too for posterity.

PR #172, opened 2016-11-09
Hcf 1057 properties per bosh release part 2

ericpromislow:
5 main changes here:

1. If package-building fails, the error is sometimes lost in an async channel. If we have a sync
error, check the channel and show the underlying error first.

2. GenerateConfigurationBase no longer used

3. In tests, put all hashes in named constants and add a comment pointing to their origin.  Yes, you
can use grep to find it, but the comment helps understand commit changes.

4. Abstract the distinguisher for multi-release properties.  Was the release pointer, is currently
just the release name, but we tried <release name>:<job name> and found it was wrong.  The extra
function doesn't currently change anything, just gives it better flexibility

5. Most importantly, spec merging is fixed by copying subtrees instead of assigning them by
reference -- since lower nodes can be modified in the destination, this corrupted the original
source tree and changed default spec  values.


test-assets/ntp-release/config/dev.yml
```diff
@@ -1,2 +1,3 @@
 ---
-latest_release_filename: "/root/hcf/test_releases/ntp-release/dev_releases/ntp/ntp-2+dev.2.yml"
+latest_release_filename: "/home/ericp/lab/gostuff/src/github.com/hpcloud/fissile/test-assets/ntp-release/dev_releases/ntp/ntp-2+dev.4.yml"
```
alexjh: I don't think this should have `ericp` in the path.

util/json_helper.go
```diff
@@ -71,39 +71,99 @@ func JSONMergeBlobs(dest, src map[string]interface{}) error {
...
+			}
+			if destValue == nil {
+				// Copy the src value for the dest
+				err := assignBlob(dest, srcKey, srcValue)
+				if err != nil {
+					return err
 				}
-				return fmt.Errorf("Invalid merge near %s: cannot merge %v with %v",
-					strings.Join(append(path, k), "."), old, v)
+				// dest[srcKey] = srcValue
```
alexjh: Is this commented out code (if so, delete) or a descriptive comment (expand on it so it doesn't look like commented code)?

util/json_helper.go
```diff
@@ -71,39 +71,99 @@ func JSONMergeBlobs(dest, src map[string]interface{}) error {
...
-			dest[k] = v
+			err := typeCheck(destValue, srcValue, srcKey, path)
+			if err != nil {
+				return err
+			}
+			err = assignBlob(dest, srcKey, srcValue)
+			if err != nil {
+				return err
+			}
+			// dest[srcKey] = srcValue
```
alexjh: Same as previous comment.

test-assets/ntp-release/config/dev.yml
ericpromislow: Agreed. Committed in 115ae0e . Note that these errors are common (path is localized when you
recreate a bosh release) but inconsequential, as the tests don't use this value.

util/json_helper.go
ericpromislow: It was old code.  assignBlob replaces `dest[srcKey] = srcValue` with `dest[srcKey] = deepcopy(srcValue)`  .  Removed in 35e2f19

util/json_helper.go
ericpromislow: See above.

aarondl:
This PR will be recreated on top of HCF-1124

PR #173, opened 2016-11-09
HCF-1112 Stream tar to docker when building role base image

Mook-as:
This is instead of using a temporary directory for the job.

This is only the first part of HCF-1112, but it's relatively independent.


builder/base_image.go
```diff
@@ -25,53 +26,89 @@ func NewBaseImageBuilder(baseImage string) *BaseImageBuilder {
...
+			defer pipeWriter.Close()
+			tarStream := tar.NewWriter(pipeWriter)
+			defer tarStream.Close()
+
+			// Generate dockerfile
+			dockerfileContents, err := b.generateDockerfile()
+			if err != nil {
+				return err
+			}
+			err = util.WriteToTarStream(tarStream, dockerfileContents, tar.Header{
```
aarondl: Is it possible we could group these together and just loop over the things we want to put in here? As opposed to 3-4 calls to WriteToTarStream etc.

builder/base_image.go
```diff
@@ -25,53 +26,89 @@ func NewBaseImageBuilder(baseImage string) *BaseImageBuilder {
...
+				Name: "Dockerfile",
+			})
+			if err != nil {
+				return err
+			}
+
+			// Add configgin
+			// The local function is to ensure we scope everything with access
+			// to the (large) configgin binary so it can be freed early
+			err = func() error {
```
aarondl: Not necessary if you just do this last in the function.

builder/base_image.go
```diff
@@ -25,53 +26,89 @@ func NewBaseImageBuilder(baseImage string) *BaseImageBuilder {
...
-		return err
-	}
-
-	if err := extractor.NewTgz().Extract(configginTarballPath, configginDir); err != nil {
-		return err
-	}
-
-	return nil
+// CreateDockerStream generates a Dockerfile and assets in the targetDir
+func (b *BaseImageBuilder) CreateDockerStream(configginTarballPath string) (io.ReadCloser, <-chan error) {
```
aarondl: As we discussed, it's probably easier to create an object that has a `.Error()` rather than have the <- chan error. The error bit would need to be protected by a mutex too though. Which opens it up for racing a little bit.

builder/base_image.go
Mook-as: `Dockerfile` is completely generated in-memory, so not really.  We might be able to do so for `monitrc.erb` and `rsyslog_conf/`, but that seems more complicated than it's worth to me?

builder/base_image.go
Mook-as: Good idea, done in 87faac06c8bd46a0411d2a708f3463cc56aa61f1

builder/base_image.go
Mook-as: From further discussion, we made it all the docker wrapper layer's job, and pass in a `func (*tar.Writer) error` instead.

builder/packages_image_test.go
```diff
@@ -96,14 +96,19 @@ func TestCreatePackagesDockerStream(t *testing.T) {
 	packagesImageBuilder, err := NewPackagesImageBuilder("foo", compiledPackagesDir, targetPath, "3.14.15", ui)
 	assert.NoError(err)
 
-	tarStream, errors, err := packagesImageBuilder.CreatePackagesDockerStream(
+	tarFile, err := ioutil.TempFile("", "fissile-test-package-image")
```
aarondl: Do we need a temp file here? Bytes buffer?

app/fissile.go
```diff
@@ -351,14 +334,15 @@ func (f *Fissile) GeneratePackagesRoleImage(repository string, roleManifest *mod
 		docker.ColoredBuildStringFunc(packagesLayerImageName),
 	)
 
-	err = dockerManager.BuildImageFromStream(tarStream, packagesLayerImageName, stdoutWriter)
+	tarPopulator := packagesImageBuilder.PopulateTarStream(roleManifest, lightManifestPath, darkManifestPath, force)
```
Mook-as: Yep, I didn't like that name. So I went for the [xkcd 386](https://www.xkcd.com/386/) way of getting a better one :grinning:

builder/base_image.go
```diff
@@ -25,53 +26,75 @@ func NewBaseImageBuilder(baseImage string) *BaseImageBuilder {
...
+			err = util.WriteToTarStream(tarWriter, assetContents, tar.Header{
+				Name: assetName,
+			})
+			if err != nil {
+				return err
+			}
+		}
+
+		// Add configgin
+		configginGzip, err := ioutil.ReadFile(configginTarballPath)
```
Mook-as: Yeah, this _will_ go away (that's the actual point of this task), so I'm just being lazy here.  Though you're right, I should just have had a reader. Whatever :p

docker/docker.go
```diff
@@ -168,13 +169,20 @@ func (d *ImageManager) BuildImage(dockerfileDirPath, name string, stdoutWriter i
 	return nil
 }
 
-// BuildImageFromStream builds a docker image using a tar file stream containing a Dockerfile
-// A io.Writer can be given for log messages; if it implments io.Closer, it will be closed when done.
-func (d *ImageManager) BuildImageFromStream(stream io.Reader, name string, stdoutWriter io.Writer) error {
+// BuildImageFromCallback builds a docker image by letting a callback popuplate
+// a tar.Writer; the callback must write a Dockerfile into the tar stream (as
+// well as any additional build context).  If stdoutWriter implements io.Closer,
+// it well be closed when done.
```
Mook-as: :paperclip: (Will fix)

builder/packages_image_test.go
Mook-as: Good call, that works (since we write exactly once and read exactly once).

PR #175, opened 2016-11-10
[HCF-1100] Added metrics for package compilation.

andreas-kupries:


Mook-as:
Sorry, my local concourse accidentally stomped over your PR checks. It was falling over for not finding `stampy`.

andreas-kupries:
Right. I should vendor `stampy` in, right ?

andreas-kupries:
Addressed with commit d6068055b179be5901ccba2a8bd6baa62b618e79

compilator/compilator_test.go
```diff
@@ -60,9 +65,18 @@ func TestCompilationEmpty(t *testing.T) {
...
+
+	contents, err := ioutil.ReadFile(metrics)
+	assert.Nil(err)
+	assert.Regexp(regexp.MustCompile(expected), string(contents))
 }
 
 func TestCompilationBasic(t *testing.T) {
+	metrics := "test-metrics.csv"
+	expected := `.*,compilator,start\n.*,compilator::job::test-release/ruby-2.5,start\n.*,compilator::job::wait::test-release/ruby-2.5,start\n.*,compilator::job::wait::test-release/ruby-2.5,done\n.*,compilator::job::run::test-release/ruby-2.5,start\n.*,compilator::job::run::test-release/ruby-2.5,done\n.*,compilator::job::test-release/ruby-2.5,done\n.*,compilator::job::test-release/go-1.4,start\n.*,compilator::job::wait::test-release/go-1.4,start\n.*,compilator::job::wait::test-release/go-1.4,done\n.*,compilator::job::run::test-release/go-1.4,start\n.*,compilator::job::run::test-release/go-1.4,done\n.*,compilator::job::test-release/go-1.4,done\n.*,compilator::job::test-release/consul,start\n.*,compilator::job::wait::test-release/consul,start\n.*,compilator::job::wait::test-release/consul,done\n.*,compilator::job::run::test-release/consul,start\n.*,compilator::job::run::test-release/consul,done\n.*,compilator::job::test-release/consul,done
+.*/compilator.go:.*,compilator,done`
```
Mook-as: Really not a fan of the giant line. Thinking of something more like
```go
expected := []string{
	`.*,compilator,start`,
	…
}
actual := strings.Split(strings.TrimSpace(string(contents)), "\n")
if assert.Len(actual, len(expected)) {
	for lineno, expectedLine := range expected {
		assert.Regexp(regexp.MustCompile(expectedLine), actual[lineno])
	}
}
```
(Though it looks like if we match line-by-line we can use `string.HasSuffix` instead of doing regexps)

Also, is line 77 supposed to be off by itself like that?

compilator/compilator_test.go
```diff
@@ -49,7 +50,11 @@ func TestMain(m *testing.M) {
 func TestCompilationEmpty(t *testing.T) {
 	assert := assert.New(t)
 
-	c, err := NewCompilator(nil, "", "", "", "", false, ui)
+	metrics := "test-metrics.csv"
```
Mook-as: Just dropping the file in the working directory is kind of bad :| Make a tempfile?

compilator/compilator_test.go
```diff
@@ -100,6 +114,10 @@ func TestCompilationBasic(t *testing.T) {
 	case <-time.After(1 * time.Second):
 		assert.Fail("Timed out waiting for overall completion")
 	}
+
+	contents, err := ioutil.ReadFile(metrics)
+	assert.Nil(err)
```
Mook-as: Use `assert.NoError` when testing against errors (it gives a better message on failure).

cmd/root.go
```diff
@@ -217,6 +225,10 @@ func extendPathsFromWorkDirectory() {
 	if flagDarkOpinions == "" {
 		flagDarkOpinions = filepath.Join(workDir, "dark-opinions.yml")
 	}
+
+	if flagMetrics == "" {
```
viovanov: I thought we don't do anything if metrics are not specified.
You have a check for this later in the code.

compilator/compilator.go
```diff
@@ -249,6 +275,10 @@ func (j compileJob) Run() {
 					color.MagentaString(j.pkg.Release.Name),
 					color.MagentaString(j.pkg.Name))
 				j.doneCh <- compileResult{pkg: j.pkg, err: errWorkerAbort}
+
+				if c.MetricsPath != "" {
+					stampy.Stamp(c.MetricsPath, "fissile", waitSeriesName, "done")
```
mook-as: Sorry I didn't spot this before:
Does this mean the wait series `done` event won't happen until the other side has finished reading your compileResult? (I don't think `j.doneCh` is buffered).

PR #178, opened 2016-11-16
HCF-1099 image metrics

jandubois:


builder/role_image.go
```diff
@@ -321,6 +324,13 @@ func (j roleBuildJob) Run() {
 		}
 	}
 
+	if j.builder.metricsPath != "" {
```
aarondl: Is it intentional that this doesn't output metrics when it doesn't have to build anything?

jandubois:
> I'm not sure why we decided to unexport a bunch of things. Maybe it was annoying linters? It doesn't matter either way.

Mostly for consistency.  The `builder` didn't export all the names, but the `compilator` did, for unknown reasons.  Discussed at https://helion-stackato.slack.com/archives/hcf/p1479247250019826

builder/role_image_test.go
```diff
@@ -372,4 +374,43 @@ func TestBuildRoleImages(t *testing.T) {
 	)
 	assert.NoError(err)
 	assert.Empty(buildersRan, "should not have ran any builders")
+
+	// Check that we write timestamps to the metrics file
```
jandubois: Yes, but that was already done.  Should I remove it?

builder/role_image.go
jandubois: Yes, what is the point to emit timings for roles that are being skipped?  We still output overall timings for all roles: https://github.com/hpcloud/fissile/blob/HCF-1099-image-metrics/app/fissile.go#L429-L432

Do you think we need the empty timing?  They just make it harder to calculate average build times, because then you have to filter them out again...

aarondl:
Satisfied with the explanations, LGTM

PR #182, opened 2016-11-23
HCF-1132 Generate job_config.json for batch run.sh

alexjh:
Creating PR for configgin changes. ~~This isn't ready to merge until https://github.com/hpcloud/configgin/pull/49 is merged to master, at which point the submodule here can be updated.~~

### DIFF
--- a/builder/role_image.go
+++ b/builder/role_image.go
@@ -2,6 +2,7 @@ package builder
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


builder/role_image_test.go
```diff
@@ -101,8 +101,6 @@ func TestGenerateRoleImageRunScript(t *testing.T) {
 
 	runScriptContents, err := roleImageBuilder.generateRunScript(rolesManifest.Roles[0])
 	assert.NoError(err)
-	assert.Contains(string(runScriptContents), "/var/vcap/jobs-src/tor/templates/data/properties.sh.erb")
```
jandubois: I don't understang why these files are no longer referenced.

builder/role_image_test.go
```diff
@@ -120,10 +118,53 @@ func TestGenerateRoleImageRunScript(t *testing.T) {
 	runScriptContents, err = roleImageBuilder.generateRunScript(rolesManifest.Roles[1])
 	assert.NoError(err)
 	assert.NotContains(string(runScriptContents), "monit -vI")
-	assert.NotContains(string(runScriptContents), "/etc/monitrc")
```
jandubois: I don't understand why `/etc/monitrc` is now referenced by role 1.

builder/role_image_test.go
alexjh: These are now referred to in `TestGenerateRoleImageJobsConfig()` because they're in the jobs json file now.

builder/role_image.go
```diff
@@ -263,6 +284,44 @@ func (r *RoleImageBuilder) generateRunScript(role *model.Role) ([]byte, error) {
...
 }
 
+func (r *RoleImageBuilder) generateJobsConfig(role *model.Role) ([]byte, error) {
+	jobsConfig := make(map[string]map[string]interface{})
+
+	for index, job := range role.Jobs {
+		jobsConfig[job.Name] = make(map[string]interface{})
+		jobsConfig[job.Name]["base"] = fmt.Sprintf("/var/vcap/jobs-src/%s/config_spec.json", job.Name)
+
+		templates := make(map[string]string)
```
alexjh: Fixed in 4c1d14f.

builder/role_image_test.go
alexjh: I changed the `run.sh` script to check if `/etc/monitrc` exists, and chmod if it does, so it's always going to have that text now.

PR #183, opened 2016-11-24
HCF-1122 Update configgin to build on OS X

jandubois:
Also add configgin build results to `make clean` target and calls
`make dist` instead of invoking `make/package` to build configgin.


make/clean
```diff
@@ -9,3 +9,5 @@ rm -f ${GIT_ROOT}/fissile
 rm -f ${GIT_ROOT}/scripts/compilation/compilation.go
 rm -f ${GIT_ROOT}/scripts/dockerfiles/dockerfiles.go
 rm -f ${GIT_ROOT}/scripts/templates/transformations.go
+rm -r ${GIT_ROOT}/scripts/configgin/configgin.go
```
alexjh: I think this should be `-f`, not `r`

jandubois: Indeed, it should.  Fixed in 77c9d6f

alexjh:
LGTM

PR #184, opened 2016-11-24
HCF-1153 Fix artifact name to use expanded origin URL

jandubois:


jandubois:
We are not going to fix this.

PR #185, opened 2016-11-25
Hcf 1124 simplify config generator

ericpromislow:


app/fissile.go
```diff
@@ -601,6 +584,32 @@ func (f *Fissile) GetDiffConfigurationBases(releasePaths []string, cacheDir stri
...
+				}
+			}
+		}
+	}
+	if patchPropertiesJob == nil {
+		return nil
+	}
+	for _, release := range f.releases {
+		for _, job := range release.Jobs {
+			if release.Name != "hcf" && job.Name != "patch-properties" {
```
mook-as: Does `if job != patchPropertiesJob` work? (I'm actually not sure; I don't _think_ there's a copy, because `release.Jobs` is a `[]*Job`, but it needs confirming…)

app/fissile.go
```diff
@@ -646,19 +655,12 @@ func getDiffsFromReleases(releases []*model.Release) (*HashDiffs, error) {
 	for idx, release := range releases {
 		configs := release.GetUniqueConfigs()
 		for _, config := range configs {
-			key, err := configstore.BoshKeyToConsulPath(config.Name, configstore.DescriptionsStore)
-			if err != nil {
-				return nil, fmt.Errorf("Error getting config %s for release %s: %s", config.Name, release.Name, err.Error())
-			}
-			hashes[idx][key] = config.Description
+			hashes[idx][config.Name] = config.Description
```
mook-as: Isn't the config name here still scoped to the release? That is, if two releases have the same config name with different descriptions, they would clobber one another.  Not to mention using the same scope for property descriptions and defaults…

builder/packages_image.go
```diff
@@ -150,16 +149,6 @@ func (p *PackagesImageBuilder) NewDockerPopulator(roleManifest *model.RoleManife
...
-			configstore.JSONProvider,
-			lightManifestPath,
-			darkManifestPath,
-			specDir,
-		)
-
-		if err = configStore.WriteBaseConfig(roleManifest); err != nil {
-			return fmt.Errorf("Error writing base config: %s", err.Error())
-		}
 		walker := &tarWalker{stream: tarWriter, root: specDir, prefix: "specs"}
```
mook-as: The `specDir` is empty, we shouldn't create it, nor show we walk it.  This also means we need to remove it from `Dockerfile-packages`.

builder/role_image.go
```diff
@@ -168,9 +171,14 @@ func (r *RoleImageBuilder) CreateDockerfileDir(role *model.Role, baseImageName s
 
 		// Symlink spec configuration file
 		// from <ROOT_DIR>/opt/hcf/specs/<ROLE_NAME>/<JOB>.json
-		specConfigSource := filepath.Join(jsonSpecsDir, role.Name, job.Name+configstore.JobConfigFileExtension)
+		specConfigSource := filepath.Join(jsonSpecsDir, role.Name, job.Name+model.JobConfigFileExtension)
+		err = job.WriteConfigs(role, specConfigSource, r.lightOpinionsPath, r.darkOpinionsPath)
```
mook-as: There's no point in symlinks anymore, since you can just create the file at the destination directly instead (they no longer in the packages layer).

app/fissile.go
```diff
@@ -601,6 +584,32 @@ func (f *Fissile) GetDiffConfigurationBases(releasePaths []string, cacheDir stri
 	return getDiffsFromReleases(f.releases)
 }
 
+// Since each job processes only its own spec, inject the hcf-release's
+// patches-properties pseudo-job's spec into all the other jobs.
+func (f *Fissile) injectPatchPropertiesJobSpec() error {
+	var patchPropertiesJob *model.Job
+	for _, release := range f.releases {
+		if release.Name == "hcf" {
```
mook-as: Where the patch properties come from needs to be configurable, since this might _not_ be using `hcf` :disappointed: (e.g. `cf-solo` does its own thing)

model/job.go
```diff
@@ -225,6 +234,123 @@ func (j *Job) loadJobSpec() (err error) {
...
 
+// MergeSpec is used to merge temporary spec patches into each job. otherJob should only be
+// the hcf/patch-properties job.
+func (j *Job) MergeSpec(otherJob *Job) {
+	// Ignore otherJob.Name, otherJob.Description
+	if otherJob.Packages != nil {
+		j.Packages = append(j.Packages, otherJob.Packages...)
+	}
+	// Skip templates
+	// Question for Vlad: how are the patch-properties templates used?
```
mook-as: Currently, they aren't (they just exist so `bosh create-release` doesn't fall over).

model/job.go
```diff
@@ -225,6 +234,123 @@ func (j *Job) loadJobSpec() (err error) {
...
+
+	config["job"].(map[string]interface{})["name"] = role.Name
+
+	var templates []map[string]string
+	for _, roleJob := range role.Jobs {
+		templates = append(templates, map[string]string{"name": roleJob.Name})
+	}
+	config["job"].(map[string]interface{})["templates"] = templates
+
+	opinions, err := newOpinions(lightOpinionsPath, darkOpinionsPath)
```
mook-as: I wish we passed in the opinion directly, but I guess that's currently not available to the caller :|

model/job_test.go
```diff
@@ -222,3 +223,36 @@ func TestJobsSort(t *testing.T) {
 	assert.Equal(jobs[0].Name, "ccc")
 	assert.Equal(jobs[1].Name, "ddd")
 }
+
+func TestJobsProperties(t *testing.T) {
+	assert := assert.New(t)
+
+	workDir, err := os.Getwd()
+	assert.Nil(err)
```
mook-as: Prefer `assert.NoError` over `assert.Nil` when testing for errors (they give better messages on failure).

model/job_test.go
```diff
@@ -222,3 +223,36 @@ func TestJobsSort(t *testing.T) {
 	assert.Equal(jobs[0].Name, "ccc")
 	assert.Equal(jobs[1].Name, "ddd")
 }
+
+func TestJobsProperties(t *testing.T) {
```
mook-as: Were all the old tests unnecessary? It seems like you're covering a lot fewer of the edge cases here.

model/job.go
```diff
@@ -225,6 +234,123 @@ func (j *Job) loadJobSpec() (err error) {
 	return nil
 }
 
+// MergeSpec is used to merge temporary spec patches into each job. otherJob should only be
+// the hcf/patch-properties job.
+func (j *Job) MergeSpec(otherJob *Job) {
+	// Ignore otherJob.Name, otherJob.Description
+	if otherJob.Packages != nil {
+		j.Packages = append(j.Packages, otherJob.Packages...)
```
mook-as: The fact that packages and properties are shared across jobs (the new ones, at least) is not an issue, right?

What happens if some of the new packages/properties already exist in the job being modified? (This needs a comment in code to describe why it's okay)

app/fissile.go
ericpromislow: I don't see how that would happen. Each config.name indexes into a release-based slice of the `hashes`

app/fissile.go
```diff
@@ -575,6 +554,10 @@ func (f *Fissile) LoadReleases(releasePaths, releaseNames, releaseVersions []str
 	}
 
 	f.releases = releases
+	err := f.injectPatchPropertiesJobSpec()
+	if err != nil {
+		return fmt.Errorf("Error loading release information: %s", err.Error())
```
ericpromislow: Done in eba31e5

app/fissile.go
ericpromislow: Done in eba31e5

app/fissile.go
```diff
@@ -601,6 +584,32 @@ func (f *Fissile) GetDiffConfigurationBases(releasePaths []string, cacheDir stri
 	return getDiffsFromReleases(f.releases)
 }
 
+// Since each job processes only its own spec, inject the hcf-release's
+// patches-properties pseudo-job's spec into all the other jobs.
+func (f *Fissile) injectPatchPropertiesJobSpec() error {
+	var patchPropertiesJob *model.Job
+	for _, release := range f.releases {
+		if release.Name == "hcf" {
+			for _, job := range release.Jobs {
```
ericpromislow: Done in eba31e5

app/fissile.go
ericpromislow: Yes, it works. Done in eba31e5

builder/packages_image.go
ericpromislow: Done in eba31e5

model/job.go
ericpromislow: Comment added in eba31e5

model/job_test.go
ericpromislow: Right.  Let me see which of the config_store tests should be reinstated.

model/job_test.go
ericpromislow: Done in 892e6f4

model/job_test.go
ericpromislow: reinstated the relevant tests in b253c59

app/fissile_test.go
```diff
@@ -136,44 +136,44 @@ func TestDevDiffConfigurations(t *testing.T) {
...
 	if assert.Equal(4, len(hashDiffs.DeletedKeys), fmt.Sprintf("Expected 4 dropped key, got %d: %s", len(hashDiffs.DeletedKeys), hashDiffs.DeletedKeys)) {
 		sort.Strings(hashDiffs.DeletedKeys)
-		assert.Equal("/descriptions/acceptance_tests/old_key", hashDiffs.DeletedKeys[0])
-		assert.Equal("/descriptions/networks/apps", hashDiffs.DeletedKeys[1])
-		assert.Equal("/spec/cf/acceptance-tests/acceptance_tests/old_key", hashDiffs.DeletedKeys[2])
-		assert.Equal("/spec/cf/cloud_controller_ng/networks/apps", hashDiffs.DeletedKeys[3])
+		assert.Equal("acceptance_tests.old_key", hashDiffs.DeletedKeys[0])
+		assert.Equal("cf.acceptance-tests.acceptance_tests.old_key", hashDiffs.DeletedKeys[1])
+		assert.Equal("cf.cloud_controller_ng.networks.apps", hashDiffs.DeletedKeys[2])
+		assert.Equal("networks.apps", hashDiffs.DeletedKeys[3])
```
mook-as: Huh, I wonder why we didn't do
```go
assert.Equal([]string{"acceptance_tests.old_key", …}, hashDiffs.DeletedKeys)
```

builder/role_image.go
```diff
@@ -166,19 +168,12 @@ func (r *RoleImageBuilder) CreateDockerfileDir(role *model.Role, baseImageName s
...
 
-		// Symlink spec configuration file
-		// from <ROOT_DIR>/opt/hcf/specs/<ROLE_NAME>/<JOB>.json
-		specConfigSource := filepath.Join(jsonSpecsDir, role.Name, job.Name+configstore.JobConfigFileExtension)
-		// into <ROOT_DIR>/var/vcap/job-src/<JOB>/config_spec.json
-		specConfigDestination := filepath.Join(jobDir, jobConfigSpecFilename+configstore.JobConfigFileExtension)
-		// The symlink must be relative, since the path is different in the docker image
-		specConfigRelativeSource, err := filepath.Rel(filepath.Dir(specConfigDestination), specConfigSource)
+		// Write spec into <ROOT_DIR>/var/vcap/job-src/<JOB>/config_spec.json
+		specConfigDestination := filepath.Join(jobDir, jobConfigSpecFilename+model.JobConfigFileExtension)
```
mook-as: Any particular reason `jobConfigSpecFileName` doesn't include the extension (now that we only have one kind anyway)?

Yes, I realize that you just copied the old code :)

builder/role_image_test.go
```diff
@@ -239,6 +256,40 @@ func TestGenerateRoleImageDockerfileDir(t *testing.T) {
...
+	buf, err := ioutil.ReadFile(jsonPath)
+	if !assert.NoError(err, "Failed to read new_hostname/config_spec.json %s\n", jsonPath) {
+		return
+	}
+	var result map[string]interface{}
+	err = json.Unmarshal(buf, &result)
+	if !assert.NoError(err, "Error unmarshalling output") {
+		return
+	}
+	assert.Equal(0, len(result["properties"].(map[string]interface{})))
```
mook-as: `assert.Empty`? (Again, for better error messages)

builder/role_image_test.go
```diff
@@ -239,6 +256,40 @@ func TestGenerateRoleImageDockerfileDir(t *testing.T) {
...
+	var expected map[string]interface{}
+	err = json.Unmarshal([]byte(`{
+		   "tor": {
+            "client_keys": null,
+            "hashed_control_password": null,
+            "hostname": "localhost",
+            "private_key": null
+        }
+	}`), &expected)
+	assert.NoError(err, "Failed to unmarshal expected data")
```
mook-as: Shouldn't we _do_ something with `expected`? Or re-marshal the correct subset of `result` and use `assert.JSONEq` on it?

cmd/build-images.go
```diff
@@ -31,13 +32,22 @@ Before running this command, you should run ` + "`fissile build layer stemcell`"
 The images will be tagged: ` + "`<repository>-<role_name>:<SIGNATURE>`" + `.
 The SIGNATURE is based on the hashes of all jobs and packages that are included in
 the image.
+
+The --patch-properties-release flag is used to distinguish the patchProperties release/job spec
+from other specs.  At most one is allowed.  Its syntax is --patch-properties-release=<RELEASE>/<JOB>.
+For hcf: the value should be "hcf/patch-properties".
```
mook-as: We don't need to document HCF here. (If we forget, we can look at `hcf.git`'s `.fissilerc` or something.)

app/fissile.go
```diff
@@ -39,6 +40,26 @@ func NewFissileApplication(version string, ui *termui.UI) *Fissile {
 	}
 }
 
+// SetPatchPropertiesDirective saves the patch-properties release and job names, if specified.
+func (f *Fissile) SetPatchPropertiesDirective(patchPropertiesDirective string) error {
+	if patchPropertiesDirective != "" {
```
mook-as: Consider just returning early if empty. (It means I can pop the condition from the stack in my head.)

compilator/compilator_test.go
```diff
@@ -614,38 +614,38 @@ func doTestCompilePackage(t *testing.T, keepInContainer bool) {
...
 	beforeCompileContainers, err := getContainerIDs(imageName)
-	assert.Nil(err)
+	assert.NoError(err)
 
 	err = comp.compilePackage(release.Packages[0])
-	assert.Nil(err)
+	assert.NoError(err)
 	afterCompileContainers, err := getContainerIDs(imageName)
-	assert.Nil(err)
+	assert.NoError(err)
```
mook-as: Thank you for fixing _all_ of them, not just the ones you touched.

model/job.go
```diff
@@ -225,6 +234,124 @@ func (j *Job) loadJobSpec() (err error) {
 	return nil
 }
 
+// MergeSpec is used to merge temporary spec patches into each job. otherJob should only be
+// the hcf/patch-properties job.  The code assumes package and property objects are immutable,
+// as they're now being shared across jobs. Also, the results are undefined if specified packages
+// or properties already exist in the destination job.
```
mook-as: But we _do_ have things like the autoscaler where the property already exists in some jobs but not the others, right?

model/job.go
```diff
@@ -225,6 +234,124 @@ func (j *Job) loadJobSpec() (err error) {
...
+func (j *Job) MergeSpec(otherJob *Job) {
+	// Ignore otherJob.Name, otherJob.Description
+	if otherJob.Packages != nil {
+		j.Packages = append(j.Packages, otherJob.Packages...)
```
mook-as: Also long as it's of the right type, you should be able to spread a nil slice.

model/job.go
```diff
@@ -225,6 +234,124 @@ func (j *Job) loadJobSpec() (err error) {
...
+	}
+	// Skip templates -- they're only in place to keep `create-release` happy.
+	if otherJob.Properties != nil {
+		j.Properties = append(j.Properties, otherJob.Properties...)
+	}
+}
+
+// WriteConfigs merges the job's spec with the opinions and writes out json-encoding to the specified path.
```
mook-as: Wording is a bit awkward here. "… and writes out the JSON-encoded result to the specified path", perhaps? "… and writes out the result as JSON to the specified path"?

model/job.go
```diff
@@ -225,6 +234,124 @@ func (j *Job) loadJobSpec() (err error) {
...
+	}
+	if err = ioutil.WriteFile(outputPath, jobJSON, 0644); err != nil {
+		return err
+	}
+	return nil
+}
+
+// getPropertiesForJob returns the parameters for the given job, using its specs and opinions
+func (j *Job) getPropertiesForJob(opinions *opinions) (map[string]interface{}, error) {
+	props := map[string]interface{}{}
```
mook-as: `make(map[string]interface{})` seems more idiomatic for making an empty map.

model/job_test.go
```diff
@@ -222,3 +223,36 @@ func TestJobsSort(t *testing.T) {
...
+
+	workDir, err := os.Getwd()
+	assert.NoError(err)
+
+	ntpReleasePath := filepath.Join(workDir, "../test-assets/ntp-release")
+	ntpReleasePathBoshCache := filepath.Join(ntpReleasePath, "bosh-cache")
+	release, err := NewDevRelease(ntpReleasePath, "", "", ntpReleasePathBoshCache)
+	assert.NoError(err)
+
+	assert.Equal(1, len(release.Jobs))
```
mook-as: `assert.Len(release.Jobs, 1)`?

model/job_test.go
```diff
@@ -222,3 +223,36 @@ func TestJobsSort(t *testing.T) {
...
+
+	lightOpinionsPath := filepath.Join(workDir, "../test-assets/ntp-opinions/opinions.yml")
+	darkOpinionsPath := filepath.Join(workDir, "../test-assets/ntp-opinions/dark-opinions.yml")
+	opinions, err := newOpinions(lightOpinionsPath, darkOpinionsPath)
+	assert.NoError(err)
+
+	properties, err := release.Jobs[0].getPropertiesForJob(opinions)
+	assert.Equal(2, len(properties))
```
mook-as: `assert.Len(properties, 2)`

compilator/compilator_test.go
ericpromislow: Would look weird to have a mix

model/job_test.go
ericpromislow: Wouldn't that then be `assert.Len(1, release.Jobs)` ?

mook-as: Nope, the [docs](https://godoc.org/github.com/stretchr/testify/assert#Assertions.Len) say it's the thing first, and length second.  I guess they think consistency is overrated?

app/fissile.go
ericpromislow: Done in bf5772d

app/fissile.go
```diff
@@ -39,6 +40,26 @@ func NewFissileApplication(version string, ui *termui.UI) *Fissile {
...
+		msgStart := "Invalid format for --patch-properties-release flag: should be RELEASE/JOB;"
+		parts := strings.SplitN(patchPropertiesDirective, "/", 2)
+		if len(parts) <= 1 {
+			return fmt.Errorf(msgStart + " only one part is specified")
+		}
+		if len(parts[0]) == 0 {
```
ericpromislow: Done in 55da76b

app/fissile.go
```diff
@@ -601,6 +606,34 @@ func (f *Fissile) GetDiffConfigurationBases(releasePaths []string, cacheDir stri
 	return getDiffsFromReleases(f.releases)
 }
 
+// Since each job processes only its own spec, inject the hcf-release's
```
ericpromislow: Done in ae2a32b

builder/role_image.go
ericpromislow: `JobConfigFileExtension` is defined in model/job.go, used once in role_image.go  `jobConfigSpecFilename` is defined and used once in role_image.go -- I'll keep a single combined constant

ericpromislow: Done in 0083382

builder/role_image_test.go
ericpromislow: I don't think we need to do anything more. If this test fails, it's due to either a failure in the json library or in the data we've given. The error message is for the developer, who can look at the string in the code to see what's gone wrong.

cmd/build-images.go
ericpromislow: Done in 1dc128b

builder/role_image_test.go
ericpromislow: done in 62541d3

model/job.go
ericpromislow: Clarified the comment in 96d9109

model/job.go
ericpromislow: done in de743b2

model/job.go
ericpromislow: 831fa3d

model/job.go
ericpromislow: 2eaa423

model/job_test.go
ericpromislow: No, look at the docs Eric

model/job_test.go
ericpromislow: All instances of `assert.Equal(..., len(...))`  changed in 0a0f11

app/fissile.go
```diff
@@ -575,6 +577,10 @@ func (f *Fissile) LoadReleases(releasePaths, releaseNames, releaseVersions []str
 	}
 
 	f.releases = releases
+	err := f.injectPatchPropertiesJobSpec()
+	if err != nil {
```
mook-as:
```go
if err := f.injectPatchPropertiesJobSpec(); err != nil { …
```
(Don't bother fixing this, it's more for future reference)

app/fissile_test.go
```diff
@@ -112,68 +112,68 @@ func TestListProperties(t *testing.T) {
...
 	}
 	assert.NotNil(release215)
 	release224, err := model.NewDevRelease(releasePathV224, "", "", cachePath)
-	assert.Nil(err)
+	assert.NoError(err)
 	assert.NotNil(release224)
 
-	assert.Equal(11, len(release215.Packages)) // temp #
-	assert.Equal(9, len(release224.Packages))
+	assert.Len(release215.Packages, 11) // temp #
```
mook-as: The comment (which was already there) is unclear.
Don't bother touching this, though :)

PR #187, opened 2016-11-29
[HCF-1155][HCF-776] Implement support for post-start scripts.

andreas-kupries:
Collapsed the history, dropped the transient monit, as per the last comment from yesterday.
The new form with a dynamically created job checking the dependencies and starting the post-start stuff when things are up worked out very well and looks to be much simpler.

Local tests ... All working.

### DIFF
--- a/scripts/dockerfiles/Dockerfile-base
+++ b/scripts/dockerfiles/Dockerfile-base
@@ -34,6 +34,9 @@ RUN useradd -m --comment 'hcf user' vcap && \
 
 ADD monitrc.erb /opt/hcf/monitrc.erb
 
+ADD post-start.sh /opt/hcf/post-start.sh
+RUN chmod ug+x /opt/hcf/post-start.sh
+
 # Install configgin
 ADD configgin /opt/hcf/configgin/
 
--- a/scripts/dockerfiles/monitrc.erb
+++ b/scripts/dockerfiles/monitrc.erb
@@ -14,3 +14,13 @@ include /etc/monit/monitrc.d/rsyslog
 
 include /var/vcap/monit/*.monitrc
 include /var/vcap/monit/job/*.monitrc
+
+# Standard fissile job to handle post-start scripts.  Currently the
+# start command does its own dependency checking to work around issues
+# in monit. When we reach use of monit 5.15+ the dependency
+# information can shift out of the script to this job. It makes the
+# job dynamic however, i.e. it will then have to be created and added
+# by run.sh (As a zomega.monitrc file in the standard place).
+
+check file zomega path /var/vcap/monit/ready
+  start program = "/opt/hcf/post-start.sh"
--- a/scripts/dockerfiles/monitrc.erb
+++ b/scripts/dockerfiles/monitrc.erb
@@ -20,7 +20,7 @@ include /var/vcap/monit/job/*.monitrc
 # in monit. When we reach use of monit 5.15+ the dependency
 # information can shift out of the script to this job. It makes the
 # job dynamic however, i.e. it will then have to be created and added
-# by run.sh (As a zomega.monitrc file in the standard place).
+# by run.sh (As a post-start.monitrc file in the standard place).
 
-check file zomega path /var/vcap/monit/ready
+check file post-start path /var/vcap/monit/ready
   start program = "/opt/hcf/post-start.sh"
--- a/scripts/dockerfiles/post-start.sh
+++ b/scripts/dockerfiles/post-start.sh
@@ -11,7 +11,7 @@
 #
 # * `monit summary` is used to check if the __other__ jobs are up.
 #   Nothing is done while they are not up yet. We know that we are
-#   `zomega`, important to exclude ourselves from the check
+#   `post-start`, important to exclude ourselves from the check
 #
 # Doing our own dependency checking works around issues in monit.
 # This can be shifted to monit itself ('depends on') when we reach use
@@ -20,7 +20,7 @@
 (
   flock -n 9 || exit 1
 
-  notyet=$(monit summary | tail -n+3 | grep -v zomega | grep -v 'Accessible\|Running')
+  notyet=$(monit summary | tail -n+3 | grep -v post-start | grep -v 'Accessible\|Running')
   if [ -z "$notyet" ]
   then
       scripts="$(find /var/vcap/jobs/*/bin -name post-start)"
--- a/scripts/dockerfiles/post-start.sh
+++ b/scripts/dockerfiles/post-start.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+#set -e
+# Cannot use this generally. Interferes with the check via `monit summary`.
+# I.e. when things are ready the failure of grep to match aborts us.
+
+# Check for post-start scripts to run. We may not have any.
+#
+# * `flock` is used to guard against monit starting this script
+#   multiple times when our dependency checking and/or the invoked
+#   scripts are taking too long.
+#
+# * `monit summary` is used to check if the __other__ jobs are up.
+#   Nothing is done while they are not up yet. We know that we are
+#   `zomega`, important to exclude ourselves from the check
+#
+# Doing our own dependency checking works around issues in monit.
+# This can be shifted to monit itself ('depends on') when we reach use
+# of monit v5.15+ where the issues are fixed.
+
+(
+  flock -n 9 || exit 1
+
+  notyet=$(monit summary | tail -n+3 | grep -v zomega | grep -v 'Accessible\|Running')
+  if [ -z "$notyet" ]
+  then
+      scripts="$(find /var/vcap/jobs/*/bin -name post-start)"
+      set -e
+      for fname in ${scripts}
+      do
+	  echo bash $fname
+	  bash $fname
+      done
+      touch /var/vcap/monit/ready
+  fi
+) 9> /var/vcap/monit/ready.lock
--- a/scripts/dockerfiles/run.sh
+++ b/scripts/dockerfiles/run.sh
@@ -11,7 +11,7 @@ fi
 # Unmark the role. We may have this file from a previous run of the
 # role, i.e. this may be a restart. Ensure that we are not seen as
 # ready yet.
-rm -f /var/vcap/monit/ready /var/vcap/monit/ready.active
+rm -f /var/vcap/monit/ready /var/vcap/monit/ready.lock
 
 # When the container gets restarted, processes may end up with different pids
 find /run -name "*.pid" -delete
@@ -127,7 +127,7 @@ set -e
     done
     touch /var/vcap/monit/ready
   fi
-) 9> /var/vcap/monit/ready.active
+) 9> /var/vcap/monit/ready.lock
 EOF
         chmod ug+x /opt/hcf/post-start.sh
 
--- a/scripts/dockerfiles/run.sh
+++ b/scripts/dockerfiles/run.sh
@@ -90,58 +90,6 @@ done
         /var/vcap/jobs/{{ $job.Name }}/bin/run
     {{ end }}
 {{ else }}
-    # Run all the scripts called post-start, if any.
-    psp=$(find /var/vcap/jobs/*/bin -name post-start)
-
-    if [ "X$psp" != X ] ; then
-	# We have post-start scripts to run.
-
-	# We do this by adding a job to the monit configuration
-	# before invoking it. This actually trivial, we simply
-	# put a .monitrc file into the directory /var/vcap/monit/job
-	# and monit will pick it up on start.
-	#
-	# The trick is to make this new job ("PS") dependent on all
-	# the existing jobs (of the role). Because the monit we have
-	# has bugs in its dependency management our job's start
-	# command does the checking manually, parsing the output of
-	# `monit summary`.  We are also using `flock` to guard against
-	# monit starting the job multiple times (due to the checking
-	# itself or the called post-start scripts taking too long).
-	# The start command is a script we arrange to run all the
-	# post-start files. Its last action is setting the monitored
-	# marker file, completing things.
-
-
-	# Create the job script, runs all the post-start scripts
-	# found, then sets the marker.
-	cat > /opt/hcf/post-start.sh <<EOF
-#!/bin/bash
-set -e
-(
-  flock -n 9 || exit 1
-  notyet=\$(monit summary | tail -n+3 |grep -v 'Accessible\|Running'|wc -l)
-  if [ \$notyet -eq 1 ] ; then
-    for fname in ${psp} ; do
-      bash \$fname
-    done
-    touch /var/vcap/monit/ready
-  fi
-) 9> /var/vcap/monit/ready.lock
-EOF
-        chmod ug+x /opt/hcf/post-start.sh
-
-	# Create the trivial configuration file for the new job. We
-	# keep using the standard timeout of 10 seconds here. The
-	# flock in the script guards against possible multiple
-	# invokations by monit.
-	cat > /var/vcap/monit/zomega.monitrc <<EOF
-check file zomega path /var/vcap/monit/ready
-  start program = "/opt/hcf/post-start.sh"
-EOF
-
-    fi
-
     # Replace bash with monit to handle both SIGTERM and SIGINT
     exec dumb-init -- monit -vI
 {{ end }}
--- a/scripts/dockerfiles/run.sh
+++ b/scripts/dockerfiles/run.sh
@@ -8,6 +8,11 @@ EOL
 exit 0
 fi
 
+# Unmark the role. We may have this file from a previous run of the
+# role, i.e. this may be a restart. Ensure that we are not seen as
+# ready yet.
+rm -f /var/vcap/monit/ready /var/vcap/monit/ready.active
+
 # When the container gets restarted, processes may end up with different pids
 find /run -name "*.pid" -delete
 if [ -d /var/vcap/sys/run ]; then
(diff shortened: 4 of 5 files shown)


scripts/dockerfiles/run.sh
```diff
@@ -85,6 +85,52 @@ done
...
+	# - Run the post-start scripts.
+	# - Get rid of the transient monit
+	# - Mark job `ready` as ok, and start the final monit,
+	#   replacing ourselves.
+	# The last two operations are also done when we have no
+	# post-start scripts.
+
+	dumb-init -- monit -vI &
+	sleep 10
+	while [ $(monit report up) -gt 1 ] ; do
```
jandubois: The comment above says:

> Wait for all jobs (except `ready`) to be ok.

This line seems to wait for just the first job to start.

scripts/dockerfiles/monitrc.erb
```diff
@@ -14,3 +14,6 @@ include /etc/monit/monitrc.d/rsyslog
 
 include /var/vcap/monit/*.monitrc
 include /var/vcap/monit/job/*.monitrc
+
+# Standard pseudo-job. Triggered to mark the role as ready.
+check file ready path /tmp/ready
```
mook-as: We may want to give this a more obvious file name, e.g. `/var/vcap/monit/ready`

scripts/dockerfiles/run.sh
```diff
@@ -85,6 +90,67 @@ done
...
+	# - Run the post-start scripts.
+	# - Get rid of the transient monit
+	# - Mark job `ready` as ok, and start the final monit,
+	#   replacing ourselves.
+	# The last two operations are also done when we have no
+	# post-start scripts.
+
+	function monit_report_down() {
+	    # 'monit report down' requires monit 5.18+. BOSH uses monit 5.6.
+	    monit status|grep status|grep -v monitoring|grep -v 'Accessible\|Running'|wc -l
```
mook-as: Could use whitespace around operators (`|`), and possibly multiple lines.

Does `monit summary` give you output that's easier to deal with?

scripts/dockerfiles/run.sh
```diff
@@ -85,6 +90,67 @@ done
...
+	}
+
+	dumb-init -- monit -vI &
```
mook-as: You don't need `dumb-init` here; that's only useful as pid 1. It might be better to make dumb-init launch run.sh instead, actually?

scripts/dockerfiles/run.sh
```diff
@@ -85,6 +90,67 @@ done
...
     {{ end }}
 {{ else }}
+    # Run all the scripts called post-start, if any.
+    function post-start-paths()
+    {
+	declare -a fnames
+	idx=0
+	for fname in $(find /var/vcap/jobs/*/bin -name post-start) ; do
+	    fnames[$idx]=$fname
+	    idx=$((idx + 1))
```
mook-as: Since we just `echo ${fnames[*]}` at the end anyway, this might be more readable as
```bash
local fnames=""
for fname in "$(find /var/vcap/jobs/*/bin -name post-start)" ; do
    fnames="${fnames} ${fname}"
done
echo "${fnames}"
```

andreas-kupries: Ok.

scripts/dockerfiles/run.sh
andreas-kupries: Mostly yes, although it also has 2 lines at the beginning which are waste.
So it will still require a bit of `grep` to get down to what we want.

scripts/dockerfiles/monitrc.erb
mook-as: Sort of kind of both? We just [install it from Ubuntu](https://github.com/hpcloud/fissile/blob/master/scripts/dockerfiles/Dockerfile-base#L15).  The only reason I'm asking that it's not in `/tmp` is to make it easier to find (since everything else lives in `/var/vcap`).

scripts/dockerfiles/run.sh
mook-as: Right, if we had `dumb-init` as pid 1 already, we can just `exec monit` instead (only so `run.sh` doesn't need to stay alive in the steady state).

scripts/dockerfiles/monitrc.erb
andreas-kupries: Ok. The key point for me is __we install__. And like are using `/var/vcap/monit` already.
Then I see no problem to use that location also.
If there is a conflict in the future it will be there anyway.

mook-as:
Running this in a vagrant box looks fine. (Needed to write that down for myself.)

scripts/dockerfiles/monitrc.erb
viovanov: I agree with @mook-as that we should use `/var/vcap/monit`
It's where people look for things that *we* do with monit

scripts/dockerfiles/run.sh
viovanov: Is there a reason not to use monit `5.18`?
IIRC there may be some licensing stuff? But I'd like us to check first.

scripts/dockerfiles/run.sh
viovanov: I'm fine with `dumb-init` as PID 1 - assuming it will help us with things?
Also remember that we're changing the entrypoint in our SDLs.
Can you look at that before making the change? We need to understand how things will fail if users use `run.sh` directly after making the change.

scripts/dockerfiles/run.sh
```diff
@@ -85,6 +90,67 @@ done
...
+	# The last two operations are also done when we have no
+	# post-start scripts.
+
+	function monit_report_down() {
+	    # 'monit report down' requires monit 5.18+. BOSH uses monit 5.6.
+	    monit status|grep status|grep -v monitoring|grep -v 'Accessible\|Running'|wc -l
+	}
+
+	dumb-init -- monit -vI &
+	sleep 10
```
viovanov: Sleep 10? I'm not a fan of this

scripts/dockerfiles/run.sh
```diff
@@ -85,6 +90,67 @@ done
...
+	while [ $(monit_report_down) -gt 1 ] ; do
+	    echo Still waiting for jobs ...
+	    sleep 10
```
viovanov: Still not a fan :/

scripts/dockerfiles/run.sh
```diff
@@ -85,6 +90,67 @@ done
...
+	done
+
+	for fname in ${psp} ; do
+	    echo bash $fname
+	    bash $fname
+	done
+
+	# Killing the monit demon now, with all jobs running does not
+	# kill the jobs, only monit itself.
+	monit quit
```
viovanov: Why do we need to do this?
I've read all the comments but I can't tell _why_ we need the _transient_ monit.

scripts/dockerfiles/run.sh
andreas-kupries: At the time of writing I had not been aware that we were in control of monit.
I thought it was something we got from bosh.
With monit under our control I see no reason to not update.
I will check where we pull monit in, license, etc.

scripts/dockerfiles/run.sh
andreas-kupries: I had forgotten about the SDL. Thank you for brining that up. To me that is (just) another (very? compelling?) reason to __not__ change the entrypoint to `dumb-init`. At least not as part of this PR. (I am trying to keep us from widening the scope of the ticket here ;)).

scripts/dockerfiles/run.sh
andreas-kupries: Hm. Will check if we can use monit itself in some way to wait and tell instead of polling in script.

scripts/dockerfiles/run.sh
andreas-kupries: Hm. ...  Ok. Let me see if I can explain how I came here ...

The original idea was to start monit earlier, then do the waiting, the post-start scripts, and then enter a long-running sleep wrapped into an infinite loop (like `while true;do sleep 100d;done`).

Then I  came to the comment `Replace bash with monit to handle both SIGTERM and SIGINT`
and realized that this loop I intended to use likely would not handle the signals properly.

But the existing command would.

To reconcile **start monit early**, and **keep the existing monit command for signals** I came up with **make the earlier monit transient**, i.e. start it, use it to wait for the processes to be up, then kill it after running post-start and transition over to the permanent monit under `dumb-init` to handle the signals.

Hopefully this was good enough to explain my reasoning which made the code as it currently is in the PR.

scripts/dockerfiles/monitrc.erb
andreas-kupries: Addressed with commit 4556357dcda7c2ae5657636982396aeb3b0c4bc8

scripts/dockerfiles/run.sh
andreas-kupries: Oh. `monit` is pulled in via `apt-get install ... monit ...`. (See `scripts/dockerfiles/Dockerfile-base`)
So, we get whatever version Ubuntu is giving us.
And that depends on the Ubuntu version used to assemble the docker image, right ?
   * 12.04: 5.3.2
   * 14.04: 5.6-2
   * 16.04: 5.16-2
   * 16.10: 5.19.0-1
   * zesty: 5.20.0-4

Do we want to build monit on our own, to get our specific version?
Ok, according to https://mmonit.com/wiki/Monit/Installation building is trivial.

Regarding licensing the main page (https://mmonit.com/monit/) claims

```
You can redistribute Monit and/or modify Monit under the terms
of the GNU Affero General Public License (AGPL). The license
only applies if you plan to distribute Monit to third parties.
```

Talked more in the standup ...

scripts/dockerfiles/run.sh
mook-as: I'm happy with it not being the entrypoint, just that it should only ever be pid 1 (in the normal case, by `exec`ing from the shell script that _is_ pid 1).  Launching it as a sub-command from bash isn't useful because then it just doesn't do anything useful since it doesn't get the orphaned children.

andreas-kupries: Addressed with commit 6c4eba597fcaa9718ad3fbe2906c99d2555de3f9.

scripts/dockerfiles/run.sh
```diff
@@ -85,6 +90,65 @@ done
...
     {{ end }}
 {{ else }}
+    # Run all the scripts called post-start, if any.
+    function post-start-paths()
+    {
+	local fnames=""
+	for fname in "$(find /var/vcap/jobs/*/bin -name post-start)" ; do
+	    fnames="${fnames} ${fname}"
+	done
+	echo "${fnames}"
```
mook-as: Looking at it again, why do we even wrap it in a variable? Isn't this method equivalent to running `find` directly?

andreas-kupries: Huh. Possibly. Will try.

andreas-kupries:
@mook-as saw something in the documentation and had an idea I quite like.
And I :headdesk: for not seeing it myself.

The main thing to note is that __monit can do dependencies between jobs__

So, we know which per-job monitrc files are included in the main monitrc.
The third argument of all `^check` commands found there gives us the names.
Now we can dynamically create a new job which depends on all these, and whose start script we can generate as well, and this start script runs the post-start scripts of the role.

Now we can simply start monit with the extended configuration and monit does everything for us.
The moment all the regular jobs are up the our dependent job triggers and is started. At the end it generates the marker file as usual, thus it is ok for the start command exit.

And the transient monit just became superfluous, including everything about killing, waiting and re-execution.

scripts/dockerfiles/run.sh
```diff
@@ -85,6 +90,58 @@ done
         /var/vcap/jobs/{{ $job.Name }}/bin/run
     {{ end }}
 {{ else }}
+    # Run all the scripts called post-start, if any.
+    psp=$(find /var/vcap/jobs/*/bin -name post-start)
+
+    if [ "X$psp" != X ] ; then
```
mook-as: Do we need the check here, or just always write out the job and have the post-start script do nothing?

scripts/dockerfiles/run.sh
```diff
@@ -85,6 +90,58 @@ done
...
+	# Create the job script, runs all the post-start scripts
+	# found, then sets the marker.
+	cat > /opt/hcf/post-start.sh <<EOF
+#!/bin/bash
+set -e
+(
+  flock -n 9 || exit 1
+  notyet=\$(monit summary | tail -n+3 |grep -v 'Accessible\|Running'|wc -l)
+  if [ \$notyet -eq 1 ] ; then
+    for fname in ${psp} ; do
```
mook-as: If we always write out the script, then we can determin `$psp` inside the script (which means the script is constant and can be written out at image building time).

scripts/dockerfiles/run.sh
```diff
@@ -85,6 +90,58 @@ done
...
+
+	# Create the job script, runs all the post-start scripts
+	# found, then sets the marker.
+	cat > /opt/hcf/post-start.sh <<EOF
+#!/bin/bash
+set -e
+(
+  flock -n 9 || exit 1
+  notyet=\$(monit summary | tail -n+3 |grep -v 'Accessible\|Running'|wc -l)
+  if [ \$notyet -eq 1 ] ; then
```
mook-as: We can get rid of the `wc -l` and do `if test -z "${notyet}"` / `if [ -z "${notyet}" ]`.

scripts/dockerfiles/run.sh
```diff
@@ -85,6 +90,58 @@ done
...
+    for fname in ${psp} ; do
+      bash \$fname
+    done
+    touch /var/vcap/monit/ready
+  fi
+) 9> /var/vcap/monit/ready.active
```
mook-as: Calling it `ready.lock` might be clearer. Not sure, up to you.

scripts/dockerfiles/run.sh
andreas-kupries: I guess we could do that. It does offend my sense of aesthetics a bit to create something which does nothing.

scripts/dockerfiles/run.sh
andreas-kupries: I do not believe so. `-z` tests for empty string.
We test that  one (`-eq 1`) job is still waiting (which by deduction is `zomega`).
I.e. the `notyet` will never be empty.

scripts/dockerfiles/run.sh
andreas-kupries: Right. Will do.

scripts/dockerfiles/run.sh
andreas-kupries: Hm. Yes, it would be. That makes it more tempting.

scripts/dockerfiles/run.sh
andreas-kupries: Addressed with commit 0f7f8c2bb336f29362f9935bf3f19912dbc9556a

scripts/dockerfiles/run.sh
mook-as: If we drop the `wc -l`, it will be empty when everything is running.

scripts/dockerfiles/run.sh
andreas-kupries: Where would I have to modify the fissile go code to add a fixed post-start.sh ?
Notes
* The `zomega` would also be fixed and can actually go into the `monitrc.erb`.
* This can be seen as a point against, as going to monit with proper dependency testing
   will have to move it out again.
* On the third hand, having it simple now is not really an argument against future complexity.

mook-as: It would be somewhere around [`/builder`](https://github.com/hpcloud/fissile/tree/master/builder) — since this is constant for every job, likely in [base_image](https://github.com/hpcloud/fissile/blob/57766e14408c38beb9f00bcd6aff9a6f963d8b30/builder/base_image.go#L30) near where we copy in `monitrc.erb`.

scripts/dockerfiles/run.sh
andreas-kupries: But `zomega` is not running/accesiable, but waiting on all the others.

andreas-kupries: That said, we can add a `grep -v zomega` to the pipe and then my point is moot .

mook-as: Oh, right. Sorry, I totally missed that (even though you commented on it). Please ignore me, then :)

andreas-kupries: So, final proposed code would be
```
notyet=\$(monit summary | tail -n+3 | grep -v zomega | grep -v 'Accessible\|Running')
+  if [ -z "\$notyet" ] ; then ...
```

scripts/dockerfiles/run.sh
andreas-kupries: Ok. I will have a look around.

andreas-kupries: I see https://github.com/hpcloud/fissile/blob/9c25c6cea566eb5f0015013ab0d5b543d3d4e7a0/builder/base_image.go#L48 ... Would simply need an additional case. Plus change to the `make/bindata` ... Nevermind, a `post-start.sh` falls under the `*.sh` pattern.

andreas-kupries: Addressed in commit ddbb065085ed237b2696f64d4deb751ca650424c

scripts/dockerfiles/Dockerfile-base
```diff
@@ -34,6 +34,9 @@ RUN useradd -m --comment 'hcf user' vcap && \
 
 ADD monitrc.erb /opt/hcf/monitrc.erb
 
+ADD post-start.sh /opt/hcf/post-start.sh
+RUN chmod ug+x /opt/hcf/post-start.sh
```
mook-as: Any particular reason this isn't `a+x`?
(If we ever end up running `run.sh` as `vcap` it will be the wrong group, as IIRC this is owned by `root/root`.)

scripts/dockerfiles/monitrc.erb
```diff
@@ -14,3 +14,13 @@ include /etc/monit/monitrc.d/rsyslog
...
+
+# Standard fissile job to handle post-start scripts.  Currently the
+# start command does its own dependency checking to work around issues
+# in monit. When we reach use of monit 5.15+ the dependency
+# information can shift out of the script to this job. It makes the
+# job dynamic however, i.e. it will then have to be created and added
+# by run.sh (As a zomega.monitrc file in the standard place).
+
+check file zomega path /var/vcap/monit/ready
+  start program = "/opt/hcf/post-start.sh"
```
mook-as: Wouldn't it be cleaner as a new file in `/var/vcap/monit`?
(I'm curious about the name, but it doesn't matter so as-is is good.)

scripts/dockerfiles/post-start.sh
```diff
@@ -0,0 +1,35 @@
...
+# of monit v5.15+ where the issues are fixed.
+
+(
+  flock -n 9 || exit 1
+
+  notyet=$(monit summary | tail -n+3 | grep -v zomega | grep -v 'Accessible\|Running')
+  if [ -z "$notyet" ]
+  then
+      scripts="$(find /var/vcap/jobs/*/bin -name post-start)"
+      set -e
```
mook-as: Shouldn't this be set right after `notyet=`? `find` doesn't appear to return an error if nothing is found.

andreas-kupries: Is it find doing the `*` expansion, or a shell ? If a shell is involved it could fail in the expansion, I think.

scripts/dockerfiles/Dockerfile-base
andreas-kupries: I thought that world-executable might be a sec-risk/

scripts/dockerfiles/monitrc.erb
mook-as: Ah, no, I mean, we don't need to have this be inside `monitrc.erb`.  Sorry for the confusion.

scripts/dockerfiles/post-start.sh
mook-as: Ah, you're right, the `*` expansion will fail, which means it will be an invalid file name, and `find` _does_ fail on that.  While that should be rare (no jobs with binaries‽), there's still no need to fail on it. So as-is is good.

scripts/dockerfiles/monitrc.erb
andreas-kupries: Oh. I thought that you were refering to the start command.
I was already familiar with having the job in the main monitrc, and it was easy to copy over.

scripts/dockerfiles/Dockerfile-base
mook-as: They can already run anything in the shell script as themselves (there's no suid involved), though? (This is mostly academic, though — it doesn't actually change anything right now, and we can flip the permissions when we need it.)

viovanov:
Like this approach :)
Not merging - assuming you'll run tests and someone can merge after.

andreas-kupries:
Yes, @mook-as is effective PoC and merger-to-be. Currently running jenkins build and tests for integration.

andreas-kupries:
Addressed in commit 45cae26beb66c6df5f02da854d9048ff50b84bad

mook-as:
Tests passed too. Merging.

PR #192, opened 2016-12-03
HCF-1137 - spike for collecting monit data from the roles.

ericpromislow:


jandubois:
This PR has been stalled for 18 month now.  Given that we want to move away from monit eventually, it seems obsolete now.

PR #196, opened 2017-01-12
HCF-1203 Use copytruncate when rotating logs

viovanov:
Use "copytruncate" for all logs other than the ones in /var/vcap (they're already rotated that way).
Use "minsize" and "maxsize" to better control when log rotation is triggered.

### DIFF
--- a/scripts/dockerfiles/rsyslog_conf/etc/logrotate.d/rsyslog
+++ b/scripts/dockerfiles/rsyslog_conf/etc/logrotate.d/rsyslog
@@ -1,14 +1,13 @@
 /var/log/syslog
 {
 	rotate 7
-	size 5M
+	maxsize=10M
+	minsize=9M
 	missingok
 	notifempty
 	delaycompress
 	compress
-	postrotate
-		reload rsyslog >/dev/null 2>&1 || true
-	endscript
+	copytruncate
 }
 
 /var/log/mail.info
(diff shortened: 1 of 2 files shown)


scripts/dockerfiles/rsyslog_conf/etc/logrotate.d/rsyslog
```diff
@@ -1,14 +1,13 @@
 /var/log/syslog
 {
 	rotate 7
-	size 5M
+	maxsize=10M
+	minsize=9M
```
mook-as: Any particular reason for the size change?
Not that rotating at 5M made sense before…

Also, you can actually have `minsize` == `maxsize` and it does the right thing.

scripts/dockerfiles/rsyslog_conf/etc/logrotate.d/rsyslog
```diff
@@ -1,14 +1,13 @@
...
 	missingok
 	notifempty
 	delaycompress
 	compress
-	postrotate
-		reload rsyslog >/dev/null 2>&1 || true
-	endscript
+	copytruncate
```
mook-as: We should be able to `killall -HUP rsyslogd` instead of `copytruncate`, but both are documented to lose data so it doesn't matter.

scripts/dockerfiles/rsyslog_conf/etc/logrotate.d/rsyslog
viovanov: Just consistency.
I am actually skeptic about the `==`
I don't mind being explicit about it

scripts/dockerfiles/rsyslog_conf/etc/logrotate.d/rsyslog
viovanov: Yeah, the reason to go with `copytruncate` is that we've seen it work for everything else.

PR #197, opened 2017-01-13
HCF-1212 Stop all processes when a container is stopped

viovanov:
Trap SIGTERM and manually run monit stop all and wait for all processes to be stopped.
When monit is stopped, it doesn't automatically stop the processes it takes care of.

### DIFF
--- a/scripts/dockerfiles/run.sh
+++ b/scripts/dockerfiles/run.sh
@@ -1,4 +1,5 @@
-#!/bin/bash
+#!/usr/bin/dumb-init --single-child /bin/bash
+
 set -e
 
 if [[ "$1" == "--help" ]]; then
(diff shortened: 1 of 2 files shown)


scripts/dockerfiles/run.sh
```diff
@@ -90,6 +91,29 @@ done
...
+    echo "All monit processes have been stopped."
+    monit summary
+    monit quit
+  }
+
+  trap killer SIGTERM
+
+  monit -vI &
+  child=$!
+  wait "$child"
```
mook-as: Note to self: `monit -v` without the `-I` would fork a child, but then we can't `wait` on it, so it's not very useful.
(We might want to add a comment about it, I guess?)

scripts/dockerfiles/run.sh
```diff
@@ -90,6 +91,29 @@ done
...
+    # Wait for all monit services to be stopped
+    echo "Received SIGTERM. Will run 'monit stop all'."
+
+    total_services=$(monit summary | grep -c "^Process")
+
+    monit stop all
+
+    echo "Ran 'monit stop all'."
+
+    while [ $total_services != $(monit summary | grep "^Process" | grep -c "Not monitored") ] ; do
```
mook-as: Is there a reason that this isn't `test -z "$(monit summary | grep '^Process' | grep -v 'Not monitored')"`? Just seemed odd to get the count first, that's all.

PR #200, opened 2017-02-03
[HCF-1912] Use type reflection to prevent inner nodes of the dark opinions to be seen as dark.

andreas-kupries:
Example: Making "uaa.clients.cf-usb.secret" dark does not mean that "uaa.clients" is dark.
This fix is required for https://github.com/hpcloud/hcf/pull/854

### DIFF
--- a/model/job.go
+++ b/model/job.go
@@ -317,9 +317,12 @@ func (j *Job) getPropertiesForJob(opinions *opinions) (map[string]interface{}, e
 
 		darkValue, ok := getOpinionValue(darkOpinionsByString, keyPieces)
 		if ok {
-			if (darkValue == nil) ||
-				((reflect.TypeOf(darkValue).Kind() != reflect.Map) &&
-				(reflect.TypeOf(darkValue).Kind() != reflect.Array)) {
+			if darkValue == nil {
+				// Ignore dark opinions
+				continue
+			}
+			kind := reflect.TypeOf(darkValue).Kind()
+			if kind != reflect.Map && kind != reflect.Array {
 				// Ignore dark opinions
 				continue
 			}
--- a/model/job.go
+++ b/model/job.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"reflect"
 	"sort"
 
 	"github.com/pivotal-golang/archiver/extractor"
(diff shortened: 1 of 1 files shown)


model/job.go
```diff
@@ -302,10 +303,26 @@ func (j *Job) getPropertiesForJob(opinions *opinions) (map[string]interface{}, e
...
+		// The check for darkness does not only test if the
+		// presented key is found in the dark opionions, but
+		// also the type of the associated value. Excluding a
+		// key like "a.b.c.d" does not mean that "a.b.c",
+		// etc. are excluded as well. Definitely not. So,
+		// finding a key we consider it to be an excluded leaf
+		// key only when the associated value, if any is
+		// neither map nor array. When finding a map or array,
+		// or no value at all we consider the key to be an
+		// inner node which is not excluded.
```
mook-as: Don't we still want to be able to mask maps? I'm thinking the following show mask `foo.masked` but not `foo.bar`:
```yaml
foo:
   masked: {}
```
Not sure what the appropriate action is for arrays, so your option is as good as any.

model/job.go
```diff
@@ -302,10 +303,26 @@ func (j *Job) getPropertiesForJob(opinions *opinions) (map[string]interface{}, e
...
+
+		darkValue, ok := getOpinionValue(darkOpinionsByString, keyPieces)
 		if ok {
-			// Ignore dark opinions
-			continue
+			if (darkValue == nil) ||
+				((reflect.TypeOf(darkValue).Kind() != reflect.Map) &&
+				(reflect.TypeOf(darkValue).Kind() != reflect.Array)) {
+				// Ignore dark opinions
+				continue
```
mook-as: This might be slightly more readable as:
```go
if darkValue == nil {
    continue
}
kind := reflect.TypeOf(darkValue).Kind()
if kind != reflect.Map && kind != reflect.Array {
    continue
}
```

model/job.go
andreas-kupries: You can still mask that. Use
```
foo:
   masked: foo
```
For the dark ops the value does not matter, so we can use a bogus value for a map.
The changed fissile gives meaning to the type of the value now.
Simple type => This is the key exclude
Map/Array =-> Don't exclude, deeper keys will be.

model/job.go
andreas-kupries: True. Noted for monday.

andreas-kupries: Adressed in comment 945eead2953d01cc933c5e0cd37c3da1ae5891dd
This also fixes the issue with formatting.

PR #201, opened 2017-03-01
HCF-1253 trigger rebuild on patch change

alexjh:


model/roles.go
```diff
@@ -168,8 +174,57 @@ func (r *Role) GetScriptPaths() map[string]string {
...
 }
 
+// GetScriptSignatures returns the SHA1 of all of the script file names and contents
+func (r *Role) GetScriptSignatures() (string, error) {
+	hasher := sha1.New()
+
+	var scripts []string
+
+	for _, f := range r.GetScriptPaths() {
+		scripts = append(scripts, f)
```
aarondl: Use copy built in here instead of a for loop + append.

model/roles_test.go
```diff
@@ -197,6 +198,88 @@ func TestRolesSort(t *testing.T) {
...
+	firstHash, _ := refRole.GetScriptSignatures()
+
+	workDir, err := ioutil.TempDir("", "fissile-test-")
+	assert.NoError(err)
+	defer os.RemoveAll(workDir)
+	releasePath := filepath.Join(workDir, "role.yml")
+
+	scriptName := "script.sh"
+	scriptPath := filepath.Join(workDir, scriptName)
+	scriptHandle, err := os.Create(scriptPath)
```
aarondl: Probably want to use [ioutil.WriteFile](https://golang.org/pkg/io/ioutil/#WriteFile)

model/roles_test.go
```diff
@@ -197,6 +198,88 @@ func TestRolesSort(t *testing.T) {
...
+		Scripts: []string{scriptName},
+		rolesManifest: &RoleManifest{
+			manifestFilePath: releasePath,
+		},
+	}
+
+	differentPatchHash, _ := differentPatch.GetScriptSignatures()
+	assert.NotEqual(firstHash, differentPatchHash, "role hash should be dependent on patch string")
+
+	scriptHandle, err = os.Create(scriptPath)
```
aarondl: Same as above

model/roles.go
alexjh: `GetScriptPaths()` returns a map, so using a loop was the only way. Made it a bit more efficient by allocating the array ahead of time.

model/roles_test.go
alexjh: Updated in both places.

viovanov:
@alexjh can you rebase this please?

alexjh:
Rebased on master, CI checks are passing.

PR #203, opened 2017-03-09
Ability for kube to consume more information from the RM

viovanov:
This includes:
- role run information
- templates and parsing mustache information
- environment variables

Also adds the capability to export kubernetes configurations.


kube/utils.go
```diff
@@ -0,0 +1,34 @@
+package kube
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// parsePortRange converts a port range string to a starting and an ending port number
```
andreas-kupries: Should (IMHO) document the format of port ranges. Less mental work than deducing it from the code.
I.e. `port ranges` are `N` and `N-M` where N, M are 16bit unsigned integers, i.e. in the range (0-655535).

kube/service.go
```diff
@@ -0,0 +1,66 @@
+package kube
+
```
andreas-kupries: Missing a test suite `service_test.go`.

kube/namespace.go
```diff
@@ -0,0 +1,14 @@
+package kube
+
```
andreas-kupries: Missing a test suite `namespace_test.go`.

model/roles.go
```diff
@@ -32,18 +49,100 @@ type Role struct {
...
 // Roles is an array of Role*
 type Roles []*Role
 
-type configuration struct {
-	Templates map[string]string `yaml:"templates"`
+// Configuration contains information about how to configure the
+// resulting images
+type Configuration struct {
+	Templates map[string]string        `yaml:"templates"`
+	Variables []*ConfigurationVariable `yaml:"variables"`
```
andreas-kupries: Could we use `ConfigurationVariableSlice` here as the type ?

model/mustache.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"fmt"
+	"sort"
+
+	"github.com/hpcloud/fissile/mustache"
+)
+
+// GetVariablesForRole returns all the environment variables required for
+// calculating all the templates for the role
+func (r *Role) GetVariablesForRole() ([]*ConfigurationVariable, error) {
```
andreas-kupries: Could we use `ConfigurationVariableSlice` here as type?

util/json_helper_test.go
```diff
@@ -58,7 +59,7 @@ func TestJSONHelperValidInput(t *testing.T) {
 		}
 		result, err := JSONMarshal(unmarshaled)
 		if testData.errMsg != "" {
-			assert.Error(err, "Exepected test sample %s to result in an error", testData.name)
+			assert.Error(err, fmt.Sprintf("Exepected test sample %s to result in an error", testData.name))
```
andreas-kupries: Typo `Exepected` --> `Expected` (inherited from before change, without correction)

kube/utils.go
viovanov: you mean as a comment?

andreas-kupries: yes

model/mustache.go
viovanov: I don't think we use that pattern anywhere in fissile. Any reason to use it here?

model/roles.go
viovanov: I don't think we use that pattern anywhere in fissile. Any reason to use it here?

andreas-kupries: Not sure what you mean about pattern (A pattern of how we use Go?). I asked because I saw later (line 123) the definition of `type ConfigurationVariableSlice []*ConfigurationVariable` and wondered if it would make sense to use the type name instead of replicating the definition

model/mustache.go
andreas-kupries: Not sure what you mean about pattern (A pattern of how we use Go?). I asked because I saw  (roles.go, line 123) the definition of `type ConfigurationVariableSlice []*ConfigurationVariable` and wondered if it would make sense to use the type name instead of replicating the definition

model/roles.go
viovanov: my bad, didn't see that
that's what I meant about pattern, yes

PR #206, opened 2017-03-16
Expose monit admin user as env var

florindragos:
We need the admin user in the galera bootstrap job. 
The currently used "admin" user does not have permissions to restart the mysql process.


jandubois:
Stale; closing.

PR #207, opened 2017-03-21
Fixes for kubed CF

mook-as:
The port names were wrong, so the services weren't working correctly.  Also some compat fixes.


viovanov:
The `-int` is killing me
I think we should find something better - we don't actually need that for anything going forward

mook-as:
If we don't need HCP compat as of right now, we can drop it now and fix up all the references in hcf.git; I'm fine with that. @aarondl ?

mook-as:
Discussed offline, we no longer need the compat

PR #208, opened 2017-03-24
Add generic error types for validation purpose

zreigz:
This PR introduces generic error types which can be put into chain in validation methods.
It helps collect all errors during validation and display them in the end.

For example when you execute `fissile build kube`:
```
  run:    
    scaling:
      min: 1
      max: 3    
    memory: 256    
    virtual-cpus: 4    
    exposed-ports:
    - name:                  #missing name
      protocol: aaaa      #wrong protocol
      external:               # missing port
      internal: 4222   
      public: false    
    - name: nats-routes      
      protocol: TCP      
      external: 4223      
      internal: 4223      
      public: false

```
you get

```
$ fissile build kube
...
run.exposed-ports.name: Required value
run.exposed-ports[].external: Invalid value: "": invalid syntax
run.exposed-ports[].protocol: Unsupported value: "aaaa": supported values: TCP, UDP

```

### DIFF
--- a/validation/errors.go
+++ b/validation/errors.go
@@ -1,0 +1,178 @@
+package validation
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+// Error is an implementation of the 'error' interface, which represents a
+// field-level validation error.
+type Error struct {
+	Type     ErrorType
+	Field    string
+	BadValue interface{}
+	Detail   string
+}
+
+// Error implements the error interface.
+func (v *Error) Error() string {
+	return fmt.Sprintf("%s: %s", v.Field, v.ErrorBody())
+}
+
+// ErrorBody returns the error message without the field name.  This is useful
+// for building nice-looking higher-level error reporting.
+func (v *Error) ErrorBody() string {
+	var s string
+	switch v.Type {
+	case ErrorTypeRequired, ErrorTypeForbidden, ErrorTypeTooLong, ErrorTypeInternal:
+		s = fmt.Sprintf("%s", v.Type)
+	default:
+		var bad string
+		badBytes, err := json.Marshal(v.BadValue)
+		if err != nil {
+			bad = err.Error()
+		} else {
+			bad = string(badBytes)
+		}
+		s = fmt.Sprintf("%s: %s", v.Type, bad)
+	}
+	if len(v.Detail) != 0 {
+		s += fmt.Sprintf(": %s", v.Detail)
+	}
+	return s
+}
+
+// ErrorType is a machine readable value providing more detail about why
+// a field is invalid.
+type ErrorType string
+
+const (
+	// ErrorTypeNotFound is used to report failure to find a requested value
+	// (e.g. looking up an ID).  See NotFound().
+	ErrorTypeNotFound ErrorType = "FieldValueNotFound"
+	// ErrorTypeRequired is used to report required values that are not
+	// provided (e.g. empty strings, null values, or empty arrays).  See
+	// Required().
+	ErrorTypeRequired ErrorType = "FieldValueRequired"
+	// ErrorTypeDuplicate is used to report collisions of values that must be
+	// unique (e.g. unique IDs).  See Duplicate().
+	ErrorTypeDuplicate ErrorType = "FieldValueDuplicate"
+	// ErrorTypeInvalid is used to report malformed values (e.g. failed regex
+	// match, too long, out of bounds).  See Invalid().
+	ErrorTypeInvalid ErrorType = "FieldValueInvalid"
+	// ErrorTypeNotSupported is used to report unknown values for enumerated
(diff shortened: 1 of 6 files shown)


validation/errors_test.go
```diff
@@ -0,0 +1,81 @@
...
+			t.Errorf("expected Type %q, got %q", testCase.expected, err.Type)
+		}
+	}
+}
+
+func TestErrorUsefulMessage(t *testing.T) {
+	s := Invalid("foo", "bar", "deet").Error()
+	for _, part := range []string{"foo", "bar", "deet", ErrorTypeInvalid.String()} {
+		if !strings.Contains(s, part) {
+			t.Errorf("error message did not contain expected part '%v'", part)
```
mook-as: This never actually tells you what the error message is, which makes it a little hard to fix.  We already pull in [testify/assert] if you want to use that ­— it does end up including the actual and expected contents in the output. (In this case, probably using `EqualError` or `Contains`, depending on how you want to go.)

[testify/assert]: https://godoc.org/github.com/stretchr/testify/assert

validation/validation.go
```diff
@@ -0,0 +1,72 @@
+package validation
+
+import (
+	"strconv"
+
+	"fmt"
```
mook-as: `go fmt` doesn't like this (move `fmt` up to line 4 before `strconv`).  Sorry, our concourse instance doesn't seem to have the fissile pipeline public… I'll check if we can fix that.

validation/validation.go
```diff
@@ -0,0 +1,72 @@
...
+		allErrs = append(allErrs, Invalid(field, value, `must be greater than or equal to 0`))
+	}
+	return allErrs
+}
+
+// ValidatePort validates that given value is valid and in range 0 - 65535.
+func ValidatePort(port string, field string) ErrorList {
+	allErrs := ErrorList{}
+
+	portInt, err := strconv.Atoi(port)
```
mook-as: Alternatively, `strconv.ParseInt(port, 10, 16)` should do both conversion and range checking.  On the other hand, the error message is not as nice… so I guess this is actually fine.

validation/rules.go
```diff
@@ -0,0 +1,26 @@
...
+	if 1 <= port && port <= 65535 {
+		return nil
+	}
+	return fmt.Errorf(`must be between %d and %d, inclusive`, 1, 65535)
+}
+
+// IsValidProtocol tests that the argument is TCP or UDP.
+func IsValidProtocol(protocol string) error {
+	if protocol != TCP && protocol != UDP {
+		return fmt.Errorf(`must be TCP or UDP`)
```
mook-as: It would be nice to be able to fix up the case for the user, but I guess we don't have a good place to do the normalization right now. There's some in [`LoadRoleManifest`](https://github.com/hpcloud/fissile/blob/8fa6780b7c966e26f4bcbb2c76f0dcf34df6ce28/model/roles.go#L196) but that would be cleaner if we moved that around.

validation/validation_test.go
```diff
@@ -0,0 +1,47 @@
...
+		"wrong ports":    {wrongPorts, false},
+		"negative field": {negativeField, false},
+	}
+
+	for name, tc := range tests {
+		errs := ValidateRoleRun(tc.run)
+		if tc.isValid && len(errs) > 0 {
+			t.Errorf("%v: unexpected error: %v", name, errs)
+		}
+		if !tc.isValid && len(errs) == 0 {
```
mook-as: Ideally, we should be checking that it's returning the correct sorts of errors, too. E.g. if we expect a non-negative field error, but got a invalid integer parse error instead.

zreigz:
@mook-as Thanks for review. Step by step I will fix all your comments

validation/errors.go
```diff
@@ -0,0 +1,181 @@
...
+// Error is an implementation of the 'error' interface, which represents a
+// field-level validation error.
+type Error struct {
+	Type     ErrorType
+	Field    string
+	BadValue interface{}
+	Detail   string
+}
+
+var _ error = &Error{}
```
zreigz: This variable was used for debug purpose and not used constantly. To avoid compilation error "not used variable"  you can use the blank identifier (_)
For example
```
package main

import (
    "fmt" // imported and not used: "fmt"
)

func main() {
    i := 1 // i declared and not used
}
```
```
package main

import (
    _ "fmt" // no more error
)

func main() {
    i := 1 // no more error
    _ = i
}
```



Because it is not used anymore I will remove it.

validation/validation.go
zreigz: I've tried to move `fmt` before 'strconv' but didn't help

zreigz: I used goimport and now it's fixed

validation/validation_test.go
zreigz: Done

zreigz:
PTAL

zreigz:
Thanks

PR #210, opened 2017-03-27
Remove some mentions of HCF

mook-as:
We should make the name generic (right now, `fissile-compat-release`, and `fissile.*` for the BOSH properties).

I did not change /opt/hcf paths; that is expected to come later as it is more likely to have issues.


viovanov:
@mook-as I think you can change `/opt/hcf` too.
Not too many things will break afaict - and it's best to break early IMO

mook-as:
I think so, too; but I'd like to do that in a separate PR (so they're easier to test separately).

PR #212, opened 2017-04-04
Add ability to build a subset of things

mook-as:
This allows you to build only the packages required for a named selection of roles, or their images, instead of doing everything at once.

This should make it easier to CI things.

I'm introducing sub-vipers to work around issues where it gets confused if multiple subcommands use the same flag name.

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -328,7 +328,7 @@ func (f *Fissile) Compile(repository, targetPath, roleManifestPath, metricsPath 
 		return fmt.Errorf("Error creating a new compilator: %s", err.Error())
 	}
 
-	roles, err := f.selectRolesToBuild(roleManifest.Roles, roleNames)
+	roles, err := roleManifest.SelectRoles(roleNames)
 	if err != nil {
 		return fmt.Errorf("Error selecting packages to build: %s", err.Error())
 	}
@@ -449,33 +449,6 @@ func (f *Fissile) GeneratePackagesRoleImage(repository string, roleManifest *mod
 	return nil
 }
 
-func (f *Fissile) selectRolesToBuild(roles model.Roles, roleNames []string) (model.Roles, error) {
-	var results model.Roles
-
-	if len(roleNames) == 0 {
-		// No role names specified, assume all roles
-		return roles, nil
-	}
-
-	var missingRoles []string
-	roleMap := make(map[string]*model.Role, len(roles))
-	for _, role := range roles {
-		roleMap[strings.ToLower(role.Name)] = role
-	}
-	for _, roleName := range roleNames {
-		if role, ok := roleMap[strings.ToLower(roleName)]; ok {
-			results = append(results, role)
-		} else {
-			missingRoles = append(missingRoles, roleName)
-		}
-	}
-	if len(missingRoles) > 0 {
-		return nil, fmt.Errorf("Some roles are unknown: %v", missingRoles)
-	}
-
-	return results, nil
-}
-
 // GenerateRoleImages generates all role images using dev releases
 func (f *Fissile) GenerateRoleImages(targetPath, repository, metricsPath string, noBuild, force bool, roleNames []string, workerCount int, rolesManifestPath, compiledPackagesPath, lightManifestPath, darkManifestPath string) error {
 	if len(f.releases) == 0 {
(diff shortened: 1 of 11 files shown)


app/fissile_test.go
```diff
@@ -177,3 +177,60 @@ func TestDevDiffConfigurations(t *testing.T) {
 	v, ok = hashDiffs.ChangedValues["cf.bogus.key"]
 	assert.False(ok)
 }
+
+func TestFisileSelectRolesToBuild(t *testing.T) {
```
andreas-kupries: Fisile => Fissile (s vs ss)

cmd/build-packages.go
```diff
@@ -25,6 +28,8 @@ compiled once.
 `,
 	RunE: func(cmd *cobra.Command, args []string) error {
 
+		flagBuildPackagesRoles := viper.GetString("roles")
```
mook-as: I know this works (at least, I've tested it), but I have no idea _how_.  This is talking to the global viper, not the custom one on line 55...

app/fissile.go
```diff
@@ -444,8 +449,35 @@ func (f *Fissile) GeneratePackagesRoleImage(repository string, roleManifest *mod
...
+
+	if len(roleNames) == 0 {
+		// No role names specified, assume all roles
+		return roles, nil
+	}
+
+	var missingRoles []string
+	roleMap := make(map[string]*model.Role, len(roles))
+	for _, role := range roles {
+		roleMap[strings.ToLower(role.Name)] = role
```
viovanov: iirc our object model already has a map for role name to role

viovanov: yeah - https://github.com/hpcloud/fissile/blob/master/model/roles.go#L42

app/fissile.go
```diff
@@ -444,8 +449,35 @@ func (f *Fissile) GeneratePackagesRoleImage(repository string, roleManifest *mod
 	return nil
 }
 
+func (f *Fissile) selectRolesToBuild(roles model.Roles, roleNames []string) (model.Roles, error) {
```
viovanov: This function can be part of `RoleManifest`

builder/packages_image.go
```diff
@@ -136,17 +136,17 @@ func (p *PackagesImageBuilder) determinePackagesLayerBaseImage(packages model.Pa
 }
 
 // NewDockerPopulator returns a function which can populate a tar stream with the docker context to build the packages layer image with
-func (p *PackagesImageBuilder) NewDockerPopulator(roleManifest *model.RoleManifest, forceBuildAll bool) func(*tar.Writer) error {
+func (p *PackagesImageBuilder) NewDockerPopulator(roles model.Roles, forceBuildAll bool) func(*tar.Writer) error {
```
viovanov: hmm - so the built subset would have its own packages layer?
I don't think that's what we want

cmd/root.go
```diff
@@ -170,20 +170,25 @@ func init() {
 
 // initConfig reads in config file and ENV variables if set.
 func initConfig() {
+	initViper(viper.GetViper())
+}
+func initViper(v *viper.Viper) {
 	if cfgFile != "" { // enable ability to specify config file via flag
-		viper.SetConfigFile(cfgFile)
```
viovanov: What's the reason for these changes?

mook-as: So that I can have multiple commands that have a `--role` argument. Otherwise it gets confused, and always uses empty for one of them (the first one bound, I think; binding overrides any existing arguments with the same name).

builder/packages_image.go
mook-as: It needs to, otherwise we'd try to build all the packages even for one role (e.g. if we wanted gorouter, we don't need Java for UAA).

That _should_ be okay, since (for development) it should still build them based on existing package layers if they already contain the correct packages (via docker labels).

mook-as: Wait what we don't actually have that code. Let me put that in...

cmd/root.go
mook-as: Ah, we _do_ have that code, near https://github.com/mook-as/fissile/blob/ebd1ad311ea40b474688182d5ce9babd88d163e0/builder/packages_image.go#L165

PR #213, opened 2017-04-07
Newer BOSH releases use a name key, not a final_name key

viovanov:


model/dev_release.go
```diff
@@ -97,10 +97,14 @@ func (r *Release) getDefaultDevReleaseName() (ver string, err error) {
 		return "", err
 	}
 
-	if value, ok := releaseConfig["final_name"]; !ok {
-		return "", fmt.Errorf("final_name key did not exist in configuration file for release: %s", r.Path)
+	if value, ok := releaseConfig["name"]; !ok {
+		if value, ok := releaseConfig["final_name"]; !ok {
+			return "", fmt.Errorf("name or final_name key did not exist in configuration file for release: %s", r.Path)
```
mook-as: Tiny grammar nit: "did not" -> "do not", since there are now two things.

PR #215, opened 2017-04-07
Building without docker

mook-as:
This implements preliminary support for building without docker.

For compilation, it requires root privileges (to create mount namespaces and do bind mounts), used to fake out `/var/vcap` across different packages.  No special privileges are required for building images (which just spit out a tarball of the docker build context).


compilator/compilator_other.go
```diff
@@ -0,0 +1,13 @@
...
+package compilator
+
+import (
+	"fmt"
+
+	"github.com/hpcloud/fissile/model"
+)
+
+func (c *Compilator) compilePackageInChroot(pkg *model.Package) (err error) {
+	return fmt.Errorf("Compilation without docker is not supports outside Linux")
```
aarondl: supported

compilator/compilator_linux_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/hpcloud/fissile/model"
+	"github.com/hpcloud/termui"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCompilePackageInChroot(t *testing.T) {
+	assert := assert.New(t)
+
+	if os.Geteuid() != 0 {
+		//t.Skip("building without docker requires root permissions")
```
aarondl: Commented code.

compilator/compilator_linux.go
```diff
@@ -0,0 +1,85 @@
...
+	// in-memory buffer of the log
+	log := new(bytes.Buffer)
+
+	stdoutWriter := docker.NewFormattingWriter(
+		log,
+		func(line string) string {
+			return color.GreenString("compilation-%s > %s", color.MagentaString("%s", pkg.Name), color.WhiteString("%s", line))
+		},
+	)
+	stderrWriter := docker.NewFormattingWriter(
```
aarondl: Might be time to move `NewFormattingWriter` out of the docker package? Seems a bit odd is all.

compilator/compilator.go
```diff
@@ -94,6 +94,7 @@ func NewCompilator(
 		repositoryPrefix: repositoryPrefix,
 		baseType:         baseType,
 		fissileVersion:   fissileVersion,
+		compilePackage:   (*Compilator).compilePackageInDocker,
```
aarondl: Might have been a good place for an interface? This is just an odd mechanism. Works fine though.

cmd/build-images.go
```diff
@@ -72,13 +79,14 @@ from other specs.  At most one is allowed.  Its syntax is --patch-properties-rel
 			workPathCompilationDir,
 			flagLightOpinions,
 			flagDarkOpinions,
+			flagOutputDirectory,
 		)
 	},
 }
+var buildImagesViper = viper.New()
```
aarondl: This is to stop overwriting the sub command flags?

builder/role_image_test.go
```diff
@@ -210,94 +211,125 @@ func TestGenerateRoleImageDockerfileDir(t *testing.T) {
...
-	// job.MF should not be there
-	assert.Error(util.ValidatePath(filepath.ToSlash(filepath.Join(dockerfileDir, "root/var/vcap/jobs-src/tor/job.MF")), false, "job manifest file"))
-
-	// And verify the config specs are as expected
-	jsonPath := filepath.Join(dockerfileDir, "root/var/vcap/jobs-src/new_hostname/config_spec.json")
-	buf, err := ioutil.ReadFile(jsonPath)
-	if !assert.NoError(err, "Failed to read new_hostname/config_spec.json %s\n", jsonPath) {
-		return
+	// Synchronize with the gofunc to make sure it's done
+	mutex.Lock()
```
aarondl: This is a race. You can't be sure anything has started in a separate thread. Use a channel latch

```go
latch := make(chan struct{})

go func() {
   close(latch) // I'm ready!
}()

<-latch // I won't be able to go until the close has occurred.
```

builder/role_image_test.go
```diff
@@ -210,94 +211,125 @@ func TestGenerateRoleImageDockerfileDir(t *testing.T) {
...
-
-	for _, info := range []struct {
-		path  string
-		isDir bool
-		desc  string
+	const TypeMissing byte = tar.TypeCont // flag to indicate an expected missing file
+	expected := map[string]struct {
+		desc     string
+		typeflag byte // default to TypeRegA
+		keep     bool // Hold for extra examination aftare
```
aarondl: aftare -> after

builder/role_image.go
```diff
@@ -372,12 +355,27 @@ func (j roleBuildJob) Run() {
...
+				if hasImage, err := j.dockerManager.HasImage(roleImageName); err != nil {
+					return err
+				} else if hasImage {
+					j.ui.Printf("Skipping build of role image %s because it exists\n", color.YellowString(j.role.Name))
+					return nil
+				}
+			} else {
+				info, err := os.Stat(outputPath)
+				if err == nil {
+					if info.IsDir() {
```
aarondl: Is this enough? What if it's an empty directory? Failed compile or something. Just asking the question.

app/fissile.go
```diff
@@ -452,8 +461,57 @@ func (f *Fissile) GeneratePackagesRoleImage(repository string, roleManifest *mod
...
+	}
+
+	packagesLayerImageName, err := packagesImageBuilder.GetRolePackageImageName(roleManifest, roles)
+	if err != nil {
+		return fmt.Errorf("Error finding role's package name: %s", err.Error())
+	}
+	outputPath := filepath.Join(outputDirectory, fmt.Sprintf("%s.tar", packagesLayerImageName))
+
+	if !force {
+		file, err := os.Open(outputPath)
```
aarondl: You can't be using the file outside the `if !force` scope, so doesn't a stat make more sense than an open here?

app/fissile.go
```diff
@@ -468,6 +526,27 @@ func (f *Fissile) GenerateRoleImages(targetPath, repository, metricsPath string,
 		return fmt.Errorf("Error loading roles manifest: %s", err.Error())
 	}
 
+	if outputDirectory != "" {
+		err = os.MkdirAll(outputDirectory, 0755)
+		if err != nil {
+			if !os.IsExist(err) {
```
aarondl:
```go
if os.IsExist(err) {
   return fmt.Errorf("Output directory exists and is not a directory")
} else if err != nil { 
   return fmt.Errorf("Failed to create output directory")
}
```

I think this is sufficient:
https://golang.org/pkg/os/#MkdirAll
The documentation says that it will not error if the path exists and is a directory. So your only two error conditions is that it exists and it's not a directory, or your hard disk has started a chemical fire.

builder/role_image.go
```diff
@@ -71,180 +71,162 @@ func NewRoleImageBuilder(repository, compiledPackagesPath, targetPath, lightOpin
...
+			if _, ok := releaseLicensesWritten[job.Release.Name]; !ok {
+				if len(job.Release.License.Files) == 0 {
+					continue
+				}
 
-		if _, ok := releaseLicensesWritten[job.Release.Name]; !ok {
-			if len(job.Release.License.Files) == 0 {
-				continue
-			}
+				releaseDir := filepath.Join("root/opt/hcf/share/doc", job.Release.Name)
```
aarondl: Is the `hcf` dir still around?

app/fissile.go
```diff
@@ -452,8 +461,57 @@ func (f *Fissile) GeneratePackagesRoleImage(repository string, roleManifest *mod
...
+// GeneratePackagesRoleTarball builds a tarball snapshot of the build context
+// for the docker image for the packages layer where all packages are included
+func (f *Fissile) GeneratePackagesRoleTarball(repository string, roleManifest *model.RoleManifest, noBuild, force bool, roles model.Roles, outputDirectory string, packagesImageBuilder *builder.PackagesImageBuilder) error {
+	if len(f.releases) == 0 {
+		return fmt.Errorf("Releases not loaded")
+	}
+
+	packagesLayerImageName, err := packagesImageBuilder.GetRolePackageImageName(roleManifest, roles)
+	if err != nil {
+		return fmt.Errorf("Error finding role's package name: %s", err.Error())
```
mook-as: Thought you didn't need `err.Error()` for `%s` either. (It was here because it was copied old code; will fix, though)

compilator/compilator_linux.go
mook-as: Not sure where to put it, though. `util`? Putting it into the dumping ground doesn't seem much better…

app/fissile.go
mook-as: Addressed in c8cadde7ef030bce25e489a1271c1d6c24493f9

app/fissile.go
mook-as: Fixed in c8cadde7ef030bce25e489a1271c1d6c24493f99

app/fissile.go
mook-as: Done in c8cadde7ef030bce25e489a1271c1d6c24493f9

builder/role_image_test.go
mook-as: Fixed in c8cadde7ef030bce25e489a1271c1d6c24493f9

app/fissile.go
```diff
@@ -468,6 +526,27 @@ func (f *Fissile) GenerateRoleImages(targetPath, repository, metricsPath string,
...
+				return fmt.Errorf("Error creating directory %s: %s", outputDirectory, err)
+			}
+			dir, err := os.Open(outputDirectory)
+			if err != nil {
+				return fmt.Errorf("Error opening directory %s: %s", outputDirectory, err)
+			}
+			defer dir.Close()
+			info, err := dir.Stat()
+			if err != nil {
+				return fmt.Errorf("Error stat()ing directory %s: %s", outputDirectory, err)
```
mook-as: Code no longer exists in c8cadde7ef030bce25e489a1271c1d6c24493f9

builder/role_image_test.go
mook-as: Fixed (using a channel, as suggested) in c8cadde7ef030bce25e489a1271c1d6c24493f9

compilator/compilator_other.go
mook-as: Fixed in c8cadde7ef030bce25e489a1271c1d6c24493f9

aarondl:
There's two small issues with naming/docs:

* We should somewhere document that --without-docker requires `--cap-add SYS_ADMIN` to run in docker again.
* BuildInChroot is a misnomer, it doesn't actually use a chroot and just a namespace.

mook-as:
Addressed, and because I wanted to look at the (markdown) docs, had to fix the generation script 😢

aarondl:
LGTM

PR #218, opened 2017-04-14
Fix env variables going into kube pods

andreas-kupries:
Ref https://trello.com/c/MwLhXpr3/14-env-vars-are-not-detected-properly-for-pods-it-looks-like-the-pods-get-more-than-they-need

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -798,7 +798,7 @@ func (f *Fissile) GenerateKube(rolesManifestPath, outputDir, repository, registr
 	}
 
 	for _, role := range rolesManifest.Roles {
-		if isDevRole(role) {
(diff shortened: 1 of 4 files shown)


app/fissile.go
```diff
@@ -842,3 +838,13 @@ roleLoop:
 
 	return nil
 }
+
+func isDevRole(role *model.Role) bool {
```
viovanov: Can this be a function of `model.Role`?

andreas-kupries: Certainly.

andreas-kupries: Done with commit 203c573bdf5b8483c5c56ce831c49fee56a27c0c

PR #220, opened 2017-04-24
Fix the handling of "properties" which are actually part of the value for a structured (hash) property.

andreas-kupries:


andreas-kupries:
The problem I currently have with testing is that the tests use a precompiled embedded `tor bosh release`. Do we have documentation on how to update it ? Because tests would need additional properties which demonstrate the cases we are looking for and currently don't afaik.

For two, `properties.foo: 3` is AFAIK invalid syntax for a `spec`. Properties are listed by name and then have a hash with `description`, possibly `default`. Would it be correct to more precisely say that you want me to look at that `default` (an `interface{}`) as part of the parent check ?

mook-as:
That's all manually created (either via `bosh create release`, or manually playing with `tar` and things).  And yeah, the syntax wasn't meant to be correct, that was for illustrative purposes only (hopefully you got my meaning though).  That a property with an int (or other not-hash) default, and a child property that can't exist.

andreas-kupries:
Oh lovely (`That's all manually created`) not. Hopefully I can just work with `bosh create release`.
Ok. That means quite a bit more work actually.

My current data structure for the bosh properties is a `map[string]map[string][]*model.Job` keyed by property named and __stringified__ default. Lets see if I can use an `interface{}` for that. If not I will need additional structure to carry this information around.

andreas-kupries:
Partly addressed in commit 83fb93dd02ae762ff4c19cb5b6751c2fbecaf7e6.
Still have to write new unit tests for (non-)hash-valued properties.

andreas-kupries:
Tests done with 0aed989392d9967a6178c1d823d6d2ffc45be299

app/fissile.go
```diff
@@ -297,35 +298,65 @@ func (f *Fissile) collectProperties() map[string]map[string]map[string]interface
...
+
+				// Handle the property's hash flag,
+				// based on the current default for
+				// it. Note that if the default is
+				// <nil> we assume that it can be a
+				// hash. This works arounds problems
+				// in the CF spec files where the two
+				// hash-valued properties we are
+				// interested in do not have defaults.
+				// (uaa.clients, cc.quota_definitions)
```
mook-as: Hmm, any particular reason the wrapping here is so short (shorter than the lines of code it's commenting on)?

make/update-tor-release
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
+
+bosh create release --force --name tor --dir ${GIT_ROOT}/test-assets/tor-boshrelease
```
mook-as: I wonder if it would make sense to create the tar file at (test) run time? Probably not in this PR, though

app/validation.go
```diff
@@ -79,6 +91,38 @@ func checkForUndefinedBOSHProperties(label string, properties map[string]string,
...
+// chain are separated by dots.
+func checkParentsOfUndefined(p string, bosh propertyDefaults) bool {
+	at := strings.LastIndex(p, ".")
+
+	for at >= 0 {
+		// While there is a dot in the property name we have a
+		// parent to check the existence of
+
+		tail := p[at:]
+		parent := strings.TrimSuffix(p, tail)
```
andreas-kupries: Yes. I wrote it more explicit to make it more clear what I was doing (removing the tail).

model/release_test.go
```diff
@@ -317,5 +317,17 @@ func TestGetDeploymentConfig(t *testing.T) {
...
+		`not.a.hash`,
+		`tor.client_keys`,
+		`tor.hashed_control_password`,
+		`tor.hostname`,
+		`tor.private_key`,
+	}
+	for _, expected := range allExpected {
+		assert.Contains(configs, expected)
+	}
+	assert.Len(configs, 7)
```
andreas-kupries: Right, oversight

app/fissile.go
andreas-kupries: Emacs auto-formatting (ESC something-something) for the mode. Can undo.

make/update-tor-release
andreas-kupries: No idea regarding sensibility. IMHO out of scope for the ticket and PR.

model/release_test.go
andreas-kupries: Done with commit 30fa129e639b70481b2c9c8595dd1b9e8d212fdd

app/fissile.go
andreas-kupries: Done with commit 30fa129e639b70481b2c9c8595dd1b9e8d212fdd

app/validation.go
andreas-kupries: __Nothing__ done with commit 30fa129e639b70481b2c9c8595dd1b9e8d212fdd
(I really consider just using `p[at:]` to be a headache when having to work with the code later again).

mook-as: Ah. I thought the other way read better, but they're equivalent, so if you think that's clearer it's fine.

PR #223, opened 2017-04-25
Missing Readme steps

dragonchaser:
This PR adds two missing steps in the README.md which are needed to build fissile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,15 @@ It does this using just the releases, without a BOSH deployment, CPIs, or a BOSH
 agent.
 
 ## Build and Install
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,7 +7,9 @@ It does this using just the releases, without a BOSH deployment, CPIs, or a BOSH
 agent.
 
 ## Build and Install
-Building fissile needs Go 1.4 or higher. You can download it from the [Golang website](https://golang.org/doc/install)
+Building fissile needs Go 1.4 or higher and Docker.
```
mook-as: Hmm, we should double check the version requirements here; I don't think anybody has tested against 1.4 in a while.  (This does _not_ block merging the PR, though, since it's no worse than not merging it.)

PR #224, opened 2017-04-25
Allow manifest private variables

andreas-kupries:
And register fissile's internal variables for use in templates.

Ticket: https://trello.com/c/e3sNBThS/63-fissile-it-s-broken
Rebased to include PR 220.
Rebased to include PR 225.

Latest changes include work for `type: user|environment`, `internal: true|false`.
Needs another review/look now.

Rebased to resolve conflicts, clean up of history. Now ready, IMVHO.

### DIFF
--- a/kube/pod_test.go
+++ b/kube/pod_test.go
@@ -164,20 +164,28 @@ func TestPodGetEnvVars(t *testing.T) {
 	}
 
 	for _, sample := range samples {
-		defaults := map[string]string{"SOME_VAR": sample.input}
+		defaults := map[string]string{
+			"SOME_VAR": sample.input,
+			"ALL_VAR":  "placeholder",
+		}
 
 		vars, err := getEnvVars(role, defaults)
 		assert.NoError(err)
 		assert.NotEmpty(vars)
 
-		found := false
+		founda := false
+		foundb := false
 		for _, result := range vars {
 			if result.Name == "SOME_VAR" {
-				found = true
+				founda = true
 				assert.Equal(sample.expected, result.Value)
 			}
+			if result.Name == "ALL_VAR" {
+				foundb = true
+			}
 		}
-		assert.True(found, "failed to find expected variable")
+		assert.True(founda, "failed to find expected variable SOME_VAR")
+		assert.True(foundb, "failed to find expected variable ALL_VAR")
 	}
 }
 
--- a/app/validation.go
+++ b/app/validation.go
@@ -1,12 +1,13 @@
 package app
 
 import (
-	//"fmt"
-	//"strings"
+	"fmt"
+	"strings"
 
 	"github.com/hpcloud/fissile/model"
 	"github.com/hpcloud/fissile/validation"
-	//"github.com/fatih/color"
+
+	"github.com/fatih/color"
 )
 
 // validateManifestAndOpinions applies a series of checks to the role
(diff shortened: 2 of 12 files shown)


andreas-kupries:
> but I think the private variables shouldn't go in the generated yaml in k8s-land.
> After all, they're not knobs the operator are allowed to tweak (they're generated by other shell scripts).

Agreed. Will change.

> This also applies to the fissile-built-in variables, which probably means

I read this as `hide the builtin from k8s too`.

> you want to move them somewhere that kube can see too

And then you wrote the opposite. Make them visible to k8s.
So, which is it ?

mook-as:
We probably want the list of fissile-internal variables (for now, `IP_ADDRESS` and `DNS_RECORD_NAME`) in as few places as possible.  Also, that they should not end up in the yaml files generated for k8s.  So expose them somewhere that the kube (golang) package can see, but kubectl (the cli that consumes those yaml files) cannot.

andreas-kupries:
> > But I think the private variables shouldn't go in the generated yaml in k8s-land.
> > After all, they're not knobs the operator are allowed to tweak (they're generated by other shell scripts).

> Agreed. Will change.

I confused myself, thinking that private variables cannot appear in the yaml already, finding a counter-example, and then thinking if that was actually private, or not. So I made the write up below.

Please review and comment

# Classes of variables in the role manifest

 1. Regular variables (Example: GARDEN_INSECURE_DOCKER_REGISTRIES).

    * A knob for the user to twiddle.
    * Must be used in at least one property (template).
    * Must be listed in the k8s yaml files for all roles which have jobs which use properties referencing the variable.

 2. Script-defined variables (Examples: MYSQL_HCF_CLUSTER_IPS, DNS_RECORD_NAME)

    * Generated by a script (or fissile runtime).
    * Taken by configgin for use in templates.
    * Used in at least one property (template).
    * Must NOT be listed in the k8s yaml files.

 3. Script-used variable (Example: HCF_LOG_PROTOCOL)

    * A knob for the user to twiddle.
    * Not used by configgin, but by scripts
    * Not used in any property template
    * Must be listed in all k8s yaml files. Because we cannot say if the role needs it or not based on the templates, so we have to safe and assume that is is needed, in all roles.

We are currently marking class 3 variables with `private: true`, and
the other classes not at all.

I would like to propose a different scheme:

 1. Do not mark class 1.
 2. Mark class 2 with `origin: script`
 3. Mark class 3 with `used: script`

Using both marks is forbidden.

andreas-kupries:
From the chat - started implementation

(Vlad)
how about:

```
type: [user|environment]
private: [true|false]

```

`type: user` means it needs to be rendered
`private: true` means it's being used by a script

(Andreas)
So `type` denotes origin, affecting inclusion in k8s yml vs not
and `private` denotes use by configgin, i.e. use in templates.
x-ed

(Vlad)
yes

(Andreas)
I believe we agree on the meaning despite different wording
x-ed
OK. +1 on that proposal

andreas-kupries:
The various examples map as:

|Variable|Type|Private|
|---|---|---|
| HCF_LOG_PROTOCOL | user | yes |
| GARDEN_INSECURE_DOCKER_REGISTRIES | user | no |
| MYSQL_HCF_CLUSTER_IPS | environment | no |
| DNS_RECORD_NAME | environment | no |

andreas-kupries:
Comments addressed with commit dfffacc9ea8236e9ad370c18b831ac0c378900a6 (and the two before that).

kube/pod_test.go
```diff
@@ -164,20 +164,28 @@ func TestPodGetEnvVars(t *testing.T) {
...
+			"ALL_VAR": "placeholder",
+		}
 
 		vars, err := getEnvVars(role, defaults)
 		assert.NoError(err)
 		assert.NotEmpty(vars)
 
-		found := false
+		founda := false
+		foundb := false
```
mook-as: Hmm, could use better names, but I guess finding better names is hard.

test-assets/role-manifests/bosh-run-bad-cpu.yml
```diff
@@ -5,3 +5,6 @@ roles:
   run:
     memory: 1
     virtual-cpus: -2
+configuration:
+  templates:
+    properties.tor.hostname: '((IP_ADDRESS))((DNS_RECORD_NAME))'
```
mook-as: Huh, do these _have_ to be referenced now?

kube/pod_test.go
andreas-kupries: Well, we could go with `foundSome` and `foundAll` ;) I was in quick mode for these.

test-assets/role-manifests/bosh-run-bad-cpu.yml
andreas-kupries: Yes. And updating all the assets drove that home ;)

andreas-kupries: We could of course (;))add another marker flag to stop that, but then it gets just silly.

andreas-kupries:
I see I have conflicts now to recent master ... That likely is with the disabling of the validation.
Will rebase sometime today.
First I want to make another run against the HCF develop head.
If necessary HCF will get a branch which fixes leftover issues with its manifest.
Same branch could also be used for integrating the new fissile post-merge.

andreas-kupries:
SCF integration branch has started at https://github.com/hpcloud/hcf/pull/915

test-assets/role-manifests/bosh-run-bad-cpu.yml
mook-as: I think if we mark the builtins as `Internal` it all works out.  You can't (from a role manifest) create an internal environment variable, but you can do so from code just fine.

andreas-kupries: Hm. I guess so. Please run it by @viovanov however, if that truly is sensible.

andreas-kupries: Vlad agreed to the change, now in the works.

andreas-kupries: Done with commit 0fe619cf6f0f28464fc78061ec02c290940f7568
Note: Rebased, collapsed history.

PR #232, opened 2017-05-15
Make compile-without-docker detection stricter

mook-as:
We need to use the fissile-built compilation image to actually build anything; however, that image already has a `/fissile-in` directory (because that's how the script to build the image was dropped in there). So we need to actually check for `/fissile-in/var/vcap` to know if we're building in docker.


aarondl:
LGTM

PR #233, opened 2017-05-18
Extract layers

aduffeck:
This PR removes the layer building functionality from fissile which is no longer needed since the stemcell image is being provided externally.
It also removes configgin from fissile because configgin has been gemified and is provided through the stemcell image.


aduffeck:
@viovanov Nothing's lost, the history is still in the old pr. But if you force-push into a pr branch you actually lose the comments that were attached to the commits that are no longer there, no? Or did github fix that meanwhile?

mook-as:
The comments, as long as they are made as part of the PR, should remain; see https://github.com/SUSE/fissile/pull/224#discussion_r113967955-body for an example.  Unfortunately you can't get to the [commit itself](https://github.com/SUSE/fissile/commit/dfffacc9ea8236e9ad370c18b831ac0c378900a6) via the PR (need URL hacking for that).

cmd/build-packages.go
```diff
@@ -75,5 +76,12 @@ func init() {
 		"Build without docker; this may adversely affect your system.  Only supported on Linux, and requires CAP_SYS_ADMIN.",
 	)
 
+	buildPackagesCmd.PersistentFlags().StringP(
```
mook-as: We need to throw an error early if this is not supplied; otherwise the issue is hard to track down.  (It shows up as an invalid docker container name because it starts with `-`)  Bonus points for actually checking that the image exists, I guess?

PR #237, opened 2017-05-30
Use tagged fissile images

alexjh:
Move `docker-registry` and `docker-organization` to top level flags so they can be used in kube, show, and build commands.


app/fissile.go
```diff
@@ -446,7 +446,7 @@ func (f *Fissile) GeneratePackagesRoleTarball(repository string, roleManifest *m
 }
 
 // GenerateRoleImages generates all role images using dev releases
-func (f *Fissile) GenerateRoleImages(targetPath, repository, stemcellImageName, metricsPath string, noBuild, force bool, roleNames []string, workerCount int, rolesManifestPath, compiledPackagesPath, lightManifestPath, darkManifestPath, outputDirectory string) error {
+func (f *Fissile) GenerateRoleImages(targetPath, registry, organization, repository, stemcellImageName, metricsPath string, noBuild, force bool, roleNames []string, workerCount int, rolesManifestPath, compiledPackagesPath, lightManifestPath, darkManifestPath, outputDirectory string) error {
```
andreas-kupries: While out of scope for this change I do wonder if we should put the exploding number of arguments into a structure (maybe make them fields of `Fissile`) in the future.

builder/role_image_test.go
```diff
@@ -455,6 +455,8 @@ func TestBuildRoleImages(t *testing.T) {
 
 	err = roleImageBuilder.BuildRoleImages(
 		rolesManifest.Roles,
+		"test-registry.com:9000",
+		"test-organization",
```
andreas-kupries: We do not seem to have tests for
1. empty registry,
1. empty org,
1. empty registry and org

andreas-kupries: Retracting, with apologies, missed the code in `TestGetRoleDevImageName()` which does test this.

PR #239, opened 2017-05-31
Trigger rebuild for fissile version changes

andreas-kupries:
Trello: https://trello.com/c/nKVdKHnE/147-fissile-version-change-no-longer-triggers-image-rebuild

Has https://github.com/SUSE/fissile/pull/240 stacked on top, same area of fissile, tested together

PR #240, opened 2017-05-31
Trigger rebuild for opinion changes

andreas-kupries:
Trello: https://trello.com/c/IemiPFAX/105-fissile-doesn-t-detect-opinion-changes-so-it-doesn-t-rebuild-images

Stacked on https://github.com/SUSE/fissile/pull/239
because of the similarity (same area of code),  tested together.

### DIFF
--- a/model/roles.go
+++ b/model/roles.go
@@ -855,25 +855,6 @@ func (r *Role) IsDevRole() bool {
 	return false
 }
 
-// GetFileSignature returns the SHA1 of a file based on its path and contents
-func GetFileSignature(filePath string) (string, error) {
-	hasher := sha1.New()
-
-	hasher.Write([]byte(filePath))
-
-	f, err := os.Open(filePath)
-	if err != nil {
-		return "", err
-	}
-
-	if _, err := io.Copy(hasher, f); err != nil {
-		return "", err
-	}
-
-	f.Close()
-	return hex.EncodeToString(hasher.Sum(nil)), nil
-}
-
 // AggregateSignatures returns the SHA1 for a slice of strings
 func AggregateSignatures(signatures []string) string {
 	hasher := sha1.New()
--- a/model/roles.go
+++ b/model/roles.go
@@ -854,3 +854,39 @@ func (r *Role) IsDevRole() bool {
 	}
 	return false
 }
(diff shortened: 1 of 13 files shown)


model/roles.go
```diff
@@ -854,3 +854,39 @@ func (r *Role) IsDevRole() bool {
 	}
 	return false
 }
+
+// GetFileSignature returns the SHA1 of a file based on its path and contents
```
alexjh: I don't see this function being used anywhere.

model/roles.go
```diff
@@ -854,3 +854,39 @@ func (r *Role) IsDevRole() bool {
...
+func GetFileSignature(filePath string) (string, error) {
```
alexjh: If this function is being used, I think we should make it a common utility function and share it between `model/job.go`, `model/package.go`, and this file. Looks like there are similar things being done in all of them.

model/roles.go
andreas-kupries: Right. I used used in earlier revisions of the code and then it fell by the wayside.
Will remove.

model/roles.go
andreas-kupries: Mooted

model/roles.go
```diff
@@ -854,3 +854,39 @@ func (r *Role) IsDevRole() bool {
...
+	hasher := sha1.New()
+
+	hasher.Write([]byte(filePath))
+
+	f, err := os.Open(filePath)
+	if err != nil {
+		return "", err
+	}
+
```
andreas-kupries: Mooted

model/roles.go
andreas-kupries: Done with commit c1b5d3cb6eec7bd263ac9033b7a8efc8116b11eb.

model/roles.go
andreas-kupries: Done with commit c1b5d3cb6eec7bd263ac9033b7a8efc8116b11eb.

model/roles.go
andreas-kupries: Done with commit c1b5d3cb6eec7bd263ac9033b7a8efc8116b11eb.

PR #241, opened 2017-06-01
kube: switch job api namespace to batch/v1 (k8s 1.6)

mook-as:
This is for compatibility with kubernetes 1.6; otherwise things don't deploy.  For some reason we don't need to touch deployment for now.


mook-as:
Outdated PR, we can redo the work when we actually take this.

PR #242, opened 2017-06-06
Emit manifest secrets as K8s secrets, not as env vars.

andreas-kupries:
Trello https://trello.com/c/tk6NUmwT/85-3-k8s-secrets-should-be-secrets-not-env-vars

This change reworks the handling of the CVs marked secret in the manifest.

1. All manifest secrets are emitted as key/value-pairs inside of one or more K8s secrets (*)
2. When emitting the EV for CVs the secret CVs do not emit their value, but a reference to their secret+key from step 1. K8s then (should) arranges for the EV to have the value from the secret, without any intervention needed in manifest scripts and the like.

(Ad *) A K8s "secret" is a mini key/value database, i.e. a single K8s secret can hold multiple manifest secrets. Multiple K8s secrets come into play when the values from the manifest secrets together are larger than the storage limit on K8s secrets, currently 1MB (See `MaxSecretSize`). In that case the manifest secrets are split across as just as many K8s secrets needed to hold them and stay under that limit (**).

(Ad **) The secrets from the SCF manifest currently need a 244K yaml file for storage, which is base64 encoded, actual size should be around 180K, so only a single secret is needed and used.

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -720,6 +720,47 @@ func (f *Fissile) GenerateKube(rolesManifestPath, outputDir, repository, registr
 		return err
 	}
 
+	// Phase I (a): Gather the secrets
+
+	cvs := model.MakeMapOfVariables(rolesManifest)
+	for key, value := range cvs {
+		if !value.Secret {
+			delete(cvs, key)
+		}
+	}
+	// cvs now holds only the secrets.
+	thesecrets, therefs, err := kube.MakeSecrets(cvs, defaults)
+	if err != nil {
+		return err
+	}
+
+	// Phase I (b): Export the secrets
+
+	secretsDir := filepath.Join(outputDir, "secrets")
+	if err = os.MkdirAll(secretsDir, 0755); err != nil {
+		return err
+	}
+
+	for _, secret := range thesecrets {
+		outputPath := filepath.Join(secretsDir, fmt.Sprintf("%s.yml", secret.ObjectMeta.Name))
+		f.UI.Printf("Writing config %s for secret %s\n",
+			color.CyanString(outputPath),
+			color.CyanString(secret.ObjectMeta.Name),
+		)
+
+		outputFile, err := os.Create(outputPath)
+		if err != nil {
+			return err
+		}
+		defer outputFile.Close()
+
+		if err := kube.WriteYamlConfig(secret, outputFile); err != nil {
+			return err
+		}
+	}
+
+	// Phase II: Export the roles
+
 	settings := &kube.ExportSettings{
 		Defaults:        defaults,
 		Registry:        registry,
@@ -728,6 +769,7 @@ func (f *Fissile) GenerateKube(rolesManifestPath, outputDir, repository, registr
 		UseMemoryLimits: useMemoryLimits,
 		FissileVersion:  fissileVersion,
 		Opinions:        opinions,
+		Secrets:         therefs,
 	}
 
 	for _, role := range rolesManifest.Roles {
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -729,7 +729,7 @@ func (f *Fissile) GenerateKube(rolesManifestPath, outputDir, repository, registr
 		}
 	}
 	// cvs now holds only the secrets.
-	thesecrets, therefs, err := kube.MakeSecrets(cvs, defaults)
+	secrets, refs, err := kube.MakeSecrets(cvs, defaults)
 	if err != nil {
 		return err
 	}
@@ -741,7 +741,7 @@ func (f *Fissile) GenerateKube(rolesManifestPath, outputDir, repository, registr
 		return err
 	}
 
-	for _, secret := range thesecrets {
+	for _, secret := range secrets {
 		outputPath := filepath.Join(secretsDir, fmt.Sprintf("%s.yml", secret.ObjectMeta.Name))
 		f.UI.Printf("Writing config %s for secret %s\n",
 			color.CyanString(outputPath),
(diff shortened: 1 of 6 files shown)


app/fissile.go
```diff
@@ -771,12 +771,54 @@ func (f *Fissile) GenerateKube(rolesManifestPath, outputDir, repository, registr
...
+	// Phase I (a): Gather the secrets
+
+	cvs := model.MakeMapOfVariables(rolesManifest)
+	for key, value := range cvs {
+		if !value.Secret {
+			delete(cvs, key)
+		}
+	}
+	// cvs now holds only the secrets.
+	thesecrets, therefs, err := kube.MakeSecrets(cvs, defaults)
```
mook-as: I don't think we need `the` on the variable names (`secrets`, `refs` are perfectly fine by themselves).

kube/pod_test.go
```diff
@@ -183,9 +201,13 @@ func TestPodGetEnvVars(t *testing.T) {
 			if result.Name == "ALL_VAR" {
 				foundb = true
 			}
+			if result.Name == "SECRET_VAR" {
+				foundc = true
```
mook-as: We should check that the format of the secret is correct

kube/secret.go
```diff
@@ -0,0 +1,101 @@
+package kube
+
+import (
+	"fmt"
+	//	"strconv"
```
mook-as: Remove commented out import?

kube/secret.go
```diff
@@ -0,0 +1,101 @@
...
+		ObjectMeta: apiv1.ObjectMeta{
+			Name: name,
+		},
+	}
+	secret.Data = make(map[string][]byte)
+	return secret
+}
+
+// Ref is an entry in the RefMap
+type Ref struct {
```
mook-as: `SecretRef` (and below, `RefMap` → `SecretRefMap`)? Otherwise it's not obvious (from other files) this has anything to do with secrets.

kube/secret.go
```diff
@@ -0,0 +1,101 @@
...
+// and key used to store their value. Note that the key has to be
+// stored, because of the transformation at (**). Ok, not truly, but
+// then we would have to replicate the transform at the place where
+// the mapping is used. I prefer to do it only once.
+type RefMap map[string]Ref
+
+// MakeSecrets creates an array of new Secrets filled with the
+// key/value pairs from the specified map. It further returns a map
+// showing which original CV name maps to what secret+key combination.
+func MakeSecrets(secrets model.CVMap, defaults map[string]string) ([]*apiv1.Secret, RefMap, error) {
```
mook-as: Don't have to think about this now, but I think in the future we'll end up with helm charts where the secrets (and possibly other variables) are never supplied. Might need to plan for that for code we write.

kube/secret.go
```diff
@@ -0,0 +1,101 @@
...
+	var thesecrets []*apiv1.Secret
+	refs := make(map[string]Ref)
+
+	max := apiv1.MaxSecretSize
+	count := 1
+	current := NewSecret(fmt.Sprintf("secret-%d", count))
```
mook-as: `current` is a bit generic. `currentFile`, maybe?

kube/secret.go
```diff
@@ -0,0 +1,101 @@
...
+		if blen > max {
+			return nil, nil, fmt.Errorf("Secret '%s' overflows K8s limit of %d bytes",
+				key, max)
+		}
+		// The current entry's value overflows the current K8s
+		// secret. Finalize it, and open a new one to store
+		// the current entry and anything after.
+		if total+blen > max {
+			thesecrets = append(thesecrets, current)
+			count = count + 1
```
mook-as: `count++` is valid in golang.

kube/secret.go
```diff
@@ -0,0 +1,101 @@
...
+		// Here it is the keys this applies to.
+
+		skey := strings.ToLower(strings.Replace(key, "_", "-", -1))
+
+		current.Data[skey] = bytes
+		refs[key] = Ref{
+			Secret: current.ObjectMeta.Name,
+			Key:    skey,
+		}
+		total = total + blen
```
mook-as: `total += blen`

kube/secret.go
andreas-kupries: This is not really a file, even it is later dumped into one.
`currentSecret` ? `currentK8sSecret` ?

app/fissile.go
andreas-kupries: Done with commit 6f6dd6b83eae459ec80c1c9c106ec15ee7b8363a

kube/pod_test.go
andreas-kupries: Not yet done. Have to look closer

kube/secret.go
andreas-kupries: Done with commit 6f6dd6b83eae459ec80c1c9c106ec15ee7b8363a

kube/secret.go
andreas-kupries: Done with commit 6f6dd6b83eae459ec80c1c9c106ec15ee7b8363a

kube/secret.go
andreas-kupries: Ok.

kube/secret.go
andreas-kupries: Done with commit 6f6dd6b83eae459ec80c1c9c106ec15ee7b8363a

kube/secret.go
andreas-kupries: Done with commit 6f6dd6b83eae459ec80c1c9c106ec15ee7b8363a

kube/secret.go
andreas-kupries: Done with commit 6f6dd6b83eae459ec80c1c9c106ec15ee7b8363a

kube/pod_test.go
andreas-kupries: Done with commit ed40ca1f62cfb7be44e86283501bb239ce99b80c
Rebased too.

PR #245, opened 2017-06-15
Add verbose option with caching details

alexjh:
Adds a verbose option to the `fissile show release` and `fissile build packages`
commands to show details about caching.

### DIFF
--- a/compilator/compilator.go
+++ b/compilator/compilator.go
@@ -161,8 +161,8 @@ type compileResult struct {
 // - synchronizer will greedily drain the <-todoCh to starve the
 //   workers out and won't wait for the <-doneCh for the N packages it
 //   drained.
-func (c *Compilator) Compile(workerCount int, releases []*model.Release, roles model.Roles) error {
-	packages, err := c.removeCompiledPackages(c.gatherPackages(releases, roles))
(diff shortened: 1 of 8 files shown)


compilator/compilator.go
```diff
@@ -703,8 +703,14 @@ func (c *Compilator) removeCompiledPackages(packages model.Packages) (model.Pack
 
 		if compiled {
 			close(c.signalDependencies[pkg.Fingerprint])
+			if verbose {
+				c.ui.Printf("found %s in %s\n", color.YellowString(pkg.Name), pkg.GetPackageCompiledDir(c.hostWorkDir))
```
mook-as: It feels like `isPackageCompiledHarness` might want to return `GetPackageCompiledDir`, so that it can change and still have the correct report… thoughts?

mook-as:
It might also be useful to show where the _release_ is; `fissile show release` shows me two `diego`s, but I don't know where they come from.

alexjh:
> It might also be useful to show where the release is; fissile show release shows me two diegos, but I don't know where they come from.

@mook-as I think I see what's happening, there are jobs and images printed separately. Fixed in 2040e1f and d6db130, if I'm understanding your comment.

PR #249, opened 2017-06-23
Update docs for detail + clarity

mook-as:
Also regenerate the docs, moving it to `docs/generated/` so we can just `rm -rf` it in the future when we want new ones.


README.md
```diff
@@ -1,133 +1,70 @@
...
+### Prerequisites
+Building fissile needs [Go 1.7] or higher and [Docker].
+
+[Go 1.7]: https://golang.org/doc/install
+[Docker]: https://www.docker.com
 
 ### Build procedure
-Execute the following commands to compile fissile
+Fissile requires generated code using additional tools, and therefore isn't
+`go get`table.
```
aarondl: Reads like `go get table`, maybe use `go get`-able

docs/configuration.md
```diff
@@ -0,0 +1,214 @@
...
+BOSH properties and related BOSHisms.  We will be working with the example role
+manifest for NATS:
+
+```yaml
+roles:
+- name: nats                       # The name of the role
+  jobs:                            # BOSH jobs this role will have
+  - name: nats
+    release_name: nats             # The name of the BOSH release this is from
+  tags:
```
aarondl: Do we have a complete list of tags we can add to this documentation?

docs/configuration.md
```diff
@@ -0,0 +1,214 @@
...
+  - name: NATS_PASSWORD
+    description: Password for NATS
+    secret: true
+    required: true
+  - name: NATS_USER
+    description: User name for NATS
+    required: true
+```
+
+Note that the variables `DNS_RECORD_NAME` and `IP_ADDRESS` are special and are
```
aarondl: Do we have a complete list of magic variables that we have?

docs/configuration.md
```diff
@@ -0,0 +1,214 @@
...
+
+# The image to use as the stemcell; note that the specific image here may be outdated
+export FISSILE_STEMCELL="splatform/fissile-stemcell-opensuse:42.2-6.ga651b2d-28.33"
+```
+
+## Building the NATS Image
+We can now assemble all the files necessary from the information above:
+
+```bash
+$ cat > role-manifest.yml # See above for contents
```
aarondl: Replace `cat` with `vim` / `nano` probably makes more sense, with a comment:
```bash
# Fill with contents above
```

docs/configuration.md
mook-as: That was it (we only have two)

docs/configuration.md
mook-as: Nothing that's likely to be kept up to date; they're free-form.
Currently, we only have `clustered` and `dev-only` (which doesn't get exported to kube). I'd like us to drop `dev-only` instead, since we only do kube now.

README.md
mook-as: Changed 2962e63266b663258437a476a59e464ea0cffad3

docs/configuration.md
mook-as: Changed in 2962e63266b663258437a476a59e464ea0cffad3

docs/configuration.md
mook-as: Changed in 2962e63266b663258437a476a59e464ea0cffad3

PR #252, opened 2017-06-29
Create separate public and private services in kube config and helm charts

jandubois:
The public service will only include the ports marked as public; the private service continues to include all defined ports.

The public service will be called `$ROLENAME-public` while the private one is still called just `$ROLENAME`.

### DIFF
--- a/kube/deployment.go
+++ b/kube/deployment.go
@@ -8,7 +8,7 @@ import (
 	extra "k8s.io/client-go/pkg/apis/extensions/v1beta1"
 )
 
-// NewDeployment creates a Deployment for the given role, and its attached service
+// NewDeployment creates a Deployment for the given role, and its attached services
 func NewDeployment(role *model.Role, settings *ExportSettings) (*extra.Deployment, *apiv1.List, error) {
 
 	podTemplate, err := NewPodTemplate(role, settings)
--- a/kube/deployment.go
+++ b/kube/deployment.go
@@ -9,14 +9,14 @@ import (
 )
 
 // NewDeployment creates a Deployment for the given role, and its attached service
-func NewDeployment(role *model.Role, settings *ExportSettings) (*extra.Deployment, *apiv1.Service, error) {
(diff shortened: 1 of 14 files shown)


kube/service.go
```diff
@@ -67,9 +105,14 @@ func NewClusterIPService(role *model.Role, headless bool) (*apiv1.Service, error
 			}
 			service.Spec.Ports = append(service.Spec.Ports, svcPort)
 		}
-		if portDef.Public {
+		if public && portDef.Public {
```
jandubois: The `&& portDef.Public` condition is redundant here.

kube/deployment.go
```diff
@@ -9,14 +9,14 @@ import (
 )
 
 // NewDeployment creates a Deployment for the given role, and its attached service
```
mook-as: Tiny nit: doc comment needs `services` (plural) now.

kube/service.go
```diff
@@ -7,17 +7,50 @@ import (
...
+	if headless {
+		svc, err := NewClusterIPService(role, true, false)
+		if err != nil {
+			return nil, err
+		}
+		if svc != nil {
+			list.Items = append(list.Items, runtime.RawExtension{Object: svc})
+		}
+	}
+	svc, err := NewClusterIPService(role, false, false)
```
mook-as: Because `true` / `false` arguments are opaque, we could do with a comment here (`// Create private services` or something).

kube/stateful_set_test.go
```diff
@@ -156,25 +176,22 @@ func TestStatefulSetPorts(t *testing.T) {
...
+						port: 80
+						targetPort: http
 				-
-					name: https
-					port: 443
-					# targetPort must be undefined for headless services
-					targetPort: 0
+						name: https
+						port: 443
+						targetPort: https
```
mook-as: Maybe we should update the test to have a private and a public port, to ensure that the differences are noted?

Issue #254, opened 2017-07-03
a typo issue in run.sh


jianqiu:
"chmod 1730 /var/spool/cron/tabs/" 
The directory "/var/spool/cron/tabs/" is not accessible in ubuntu os, it is specific for openSuSe

PR #255, opened 2017-07-04
Test for existence of cron tabs dir before chown

alexjh:
Fixes https://github.com/SUSE/fissile/issues/254.

`/var/spool/cron/tabs/` is SUSE specific.

### DIFF
--- a/scripts/dockerfiles/run.sh
+++ b/scripts/dockerfiles/run.sh
@@ -94,7 +94,7 @@ chmod 775 /var/vcap/sys/run
 
 # Fix permissions
 chmod 640 /var/log/messages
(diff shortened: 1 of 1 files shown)


scripts/dockerfiles/run.sh
```diff
@@ -94,7 +94,10 @@ chmod 775 /var/vcap/sys/run
 
 # Fix permissions
 chmod 640 /var/log/messages
-chmod 1730 /var/spool/cron/tabs/
+if [ -e /var/spool/cron/tabs ]
```
mook-as: `-d` here (only directories), right?

alexjh: Fixed in be90547

PR #257, opened 2017-07-04
Add --create-helm-chart option to `fissile build kube`

jandubois:


app/fissile.go
```diff
@@ -835,14 +901,22 @@ func (f *Fissile) GenerateKube(rolesManifestPath, outputDir, repository, registr
...
 		Opinions:        opinions,
 		Secrets:         refs,
+		CreateHelmChart: createHelmChart,
 	}
 
 	for _, role := range rolesManifest.Roles {
 		if role.IsDevRole() {
 			continue
 		}
+		if createHelmChart && role.Run.FlightStage == model.FlightStageManual {
```
mook-as: What's the recommended way to get smoke/cats configs, if we can't helm them?

app/fissile.go
```diff
@@ -772,7 +773,7 @@ func compareHashes(v1Hash, v2Hash keyHash) *HashDiffs {
 
 // GenerateKube will create a set of configuration files suitable for deployment
 // on Kubernetes
-func (f *Fissile) GenerateKube(rolesManifestPath, outputDir, repository, registry, organization, fissileVersion string, defaultFiles []string, useMemoryLimits bool, opinions *model.Opinions) error {
+func (f *Fissile) GenerateKube(rolesManifestPath, outputDir, repository, registry, organization, fissileVersion string, defaultFiles []string, useMemoryLimits bool, createHelmChart bool, chartFilename string, opinions *model.Opinions) error {
```
mook-as: Would it make sense to squash the new params, and just say if `chartFilename` is not empty, create a chart?

kube/secret.go
```diff
@@ -50,9 +60,30 @@ func MakeSecrets(secrets model.CVMap, defaults map[string]string) ([]*apiv1.Secr
...
-		ok, strValue := configValue(value, defaults)
-		if !ok {
-			return nil, nil, fmt.Errorf("Secret '%s' has no value", key)
+		var strValue string
+		if createHelmChart {
+			if value.Generator == nil || value.Generator.Type != model.GeneratorTypePassword {
+				errString := fmt.Sprintf("A valid .Values.env.%s is required", value.Name)
+				strValue = fmt.Sprintf(`{{ required "%s" .Values.env.%s | b64enc | quote }}`, errString, value.Name)
+			} else {
+				if value.Generator.Type == model.GeneratorTypePassword {
```
mook-as: This and the `if` around might be easier as a `switch`
```go
switch {
case value.Generator == nil:
	errString := fmt.Sprintf("A valid .Values.env.%s is required", value.Name)
	strValue = fmt.Sprintf(`{{ required "%s" .Values.env.%s | b64enc | quote }}`, errString, value.Name)
case value.Generator.Type == model.GeneratorTypePassword:
	strValue = "{{ randAlphaNum 32 | b64enc | quote }}"
default:
	// no generator
}
```

app/fissile.go
jandubois: I think they probably should have their own charts. They are not really part of the deployment itself. Please file a new ticket for them.

Right now you can still run `make kube` and get the corresponding kube config file.

kube/pod.go
```diff
@@ -204,9 +204,19 @@ func getEnvVars(role *model.Role, defaults map[string]string, secrets SecretRefM
...
 		}
 
-		ok, stringifiedValue := configValue(config, defaults)
-		if !ok {
-			continue
+		var stringifiedValue string
+		if createHelmChart {
+			required := ""
+			if config.Required {
+				required = fmt.Sprintf(`required "A valid .Values.env.%s is required" `, config.Name)
```
mook-as: The error message is really unwieldy; does `A valid %s is required` not provide enough? (Why `.Values.env`?)

app/fissile.go
jandubois: Maybe, but it feels like a hack to me. Note that `--chart-file` is the name of a file (normally `Chart.yaml` that is being copied into the generated chart (directory).

kube/pod.go
jandubois: It is just following the example from the https://docs.helm.sh/developing_charts/#charts_tips_and_tricks documentation:

```
value: {{required "A valid .Values.who entry required!" .Values.who }}
```

I find the whole concept weird that both the `required` function as well as the `default` function are used during template expansion instead of being attributes in the `values.yaml` file, but I guess that is a separate topic.

The `.Values.env` part is included because that is the full scope of the variables defined in `values.yaml` right now.  If you use them in a template they would be `{{ .Values.env.UAA_PORT }}`.

kube/secret.go
jandubois: Done in 7612e9a9dcfd52a30244dd6ab9535f1f9ef87066

app/fissile.go
mook-as: Okay, leaving them separate is fine too.

app/fissile.go
mook-as: Created https://trello.com/c/hnfI8xtR/ for it.

kube/pod.go
mook-as: Alright. That message is sort of hostile to the people who actually would be running helm (who wouldn't be writing the charts, in our case), but it has _enough_ information that it's still parsable.

kube/secret.go
```diff
@@ -50,9 +60,27 @@ func MakeSecrets(secrets model.CVMap, defaults map[string]string) ([]*apiv1.Secr
 	total := 0 // Accumulated size of the values stored in 'currentSecret'
 
 	for key, value := range secrets {
-		ok, strValue := configValue(value, defaults)
-		if !ok {
-			return nil, nil, fmt.Errorf("Secret '%s' has no value", key)
+		var strValue string
+		if createHelmChart {
+			switch {
+			case value.Generator == nil || value.Generator.Type != model.GeneratorTypePassword:
```
mook-as: Why is `value.Generator.Type != model.GeneratorTypePassword` here?

kube/secret.go
```diff
@@ -50,9 +60,27 @@ func MakeSecrets(secrets model.CVMap, defaults map[string]string) ([]*apiv1.Secr
...
+				errString := fmt.Sprintf("A valid .Values.env.%s is required", value.Name)
+				strValue = fmt.Sprintf(`{{ required "%s" .Values.env.%s | b64enc | quote }}`, errString, value.Name)
+			case value.Generator.Type == model.GeneratorTypePassword:
+				strValue = "{{ randAlphaNum 32 | b64enc | quote }}"
+			default:
```
mook-as: How would this be reached?

kube/secret.go
jandubois: Because I haven't implemented any generators besides the `Password` type.  All others (certs) will be treated as configuration variables for now.

kube/secret.go
jandubois: It can't; this is leftover code from when the certs where not treated as config vars, but included literally. I'll remove it.

kube/secret.go
mook-as: Yeah, but that was what `default:` was, right?

jandubois: No, default was: Inline literal value, encoded in base64.

This is the same as what is being used for the kube config, which requires that *all* dev settings are defined.

mook-as: Right, I just meant that we could have changed `default:` to cover the `value.Generator.Type != model.GeneratorTypePassword` case instead, to make the code more readable (and maybe have a custom message).  This works too, though.

jandubois: I don't see how you can do this with `default`.  You need to cover the `value.Generator == nil` case first before you can handle `case value.Generator.Type == model.GeneratorTypePassword`. So you would then end up with duplicate code.

Lmk if I'm not seeing the obvious here!

mook-as: The [switch statement] in go is actually more like an extended `if`/`else` chain:
> In an expression switch, the switch expression is evaluated and the case expressions, which need not be constants, are evaluated left-to-right and top-to-bottom; the first one that equals the switch expression triggers execution of the statements of the associated case; the other cases are skipped.

So as long as `value.Generator == nil` comes before `case value.Generator.Type` it's fine.  So you do end up with duplicate code, but the message can then be something like `Generator type %s is not supported for .Values.env.%s, required "%s"`.

[switch statement]: https://golang.org/ref/spec#Switch_statements

jandubois: Yeah, but we don't want an error message, we want them to be treated as regular variables so users can override.

app/fissile.go
viovanov: Are we keeping the capability to create non-templatized kube configs?

app/fissile.go
```diff
@@ -824,6 +828,68 @@ func (f *Fissile) GenerateKube(rolesManifestPath, outputDir, repository, registr
 		}
 	}
 
+	if createHelmChart {
```
viovanov: Why not create a new function `GenerateHelm`?
Would be easier to read afaict.

cmd/build-kube.go
```diff
@@ -11,6 +11,8 @@ var (
 	flagBuildKubeOutputDir       string
 	flagBuildKubeDefaultEnvFiles []string
 	flagBuildKubeUseMemoryLimits bool
+	flagBuildKubeCreateHelmChart bool
```
viovanov: Why not `fissile build helm`?
This seems cluttered to me.

kube/pod.go
viovanov: Maybe I'm missing something - can't the error be `UAA_PORT configuration missing.`?
I know the full name is longer, but the operator doesn't care.

app/fissile.go
jandubois: We are so far, so just avoid breaking any workflows that rely on them. They are mostly redundant though, except for the manual bosh-tasks, which aren't generated for the helm charts.

We have a card for the tasks (https://trello.com/c/hnfI8xtR/209-helm-charts-for-smoke-cats-etc); once that is done we can drop the old kube-config (after approval from all stakeholders).

kube/pod.go
jandubois: Yes, we can make it whatever we want. But once we implement the new proposal and split the config settings into multiple namespaces, we will have to include the namespace part again.

I'll remove the prefix now since both of you seem to be bothered by it, but note that it will not improve the actual error message that much:

```
Error: render error in "uaa/templates/secret-1.yml": template: uaa/templates/secret-1.yml:10:30: executing "uaa/templates/secret-1.yml" at <required "A valid .V...>: error calling required: A valid .Values.env.UAA_ADMIN_CLIENT_SECRET is required
```

then becomes

```
Error: render error in "uaa/templates/secret-1.yml": template: uaa/templates/secret-1.yml:10:30: executing "uaa/templates/secret-1.yml" at <required "UAA_ADMIN...>: error calling required: UAA_ADMIN_CLIENT_SECRET configuration missing
```

app/fissile.go
jandubois: I kind of agree, but then the whole function (or even file) should be refactored. Do you want this done as part of this task?

cmd/build-kube.go
jandubois: It is just a variant of `fissile build kube`. Having separate command would be a lot of code duplication. But maybe that is ok, if we eventually get rid of `build kube`.

app/fissile.go
viovanov: I think keeping kube configs a little while longer is not a bad idea.
Helm might not be available everywhere yet.

app/fissile.go
viovanov: Yeah, for making future work easier.

cmd/build-kube.go
viovanov: Ok, we can have it make more sense when we complete the work

PR #259, opened 2017-07-14
Fix package layer signature calculation

mook-as:
The package layer tag was accidentally including the template information. This is causing unnecessary rebuilds.

### DIFF
--- a/builder/packages_image.go
+++ b/builder/packages_image.go
@@ -3,11 +3,14 @@ package builder
 import (
 	"archive/tar"
 	"bytes"
+	"crypto/sha1"
+	"encoding/hex"
 	"fmt"
 	"html/template"
 	"io"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


builder/packages_image.go
```diff
@@ -260,17 +263,31 @@ func (p *PackagesImageBuilder) generateDockerfile(baseImage string, packages mod
 
 // GetRolePackageImageName generates a docker image name for the amalgamation for a role image
```
andreas-kupries: `for a` --> `of a` ?

andreas-kupries: Ah. No. This is the `packages layer`. The `role` thing confused me.
Maybe change the whole function name to `GetPackageLayerImageName` ?
And the description would then be
```
GetPackageLayerImageName generates a docker image name for the amalgamation holding the packages used in all roles.
```

mook-as: Fixed in 77525b733063a37a060b509a26f72394c892eaa0 (`GetPackagesLayerImageName()`)

PR #261, opened 2017-07-14
Make storage-class names configurable via values.yaml

jandubois:


kube/stateful_set.go
```diff
@@ -52,7 +52,7 @@ func NewStatefulSet(role *model.Role, settings *ExportSettings) (*v1beta1.Statef
 }
 
 // getVolumeClaims returns the list of persistent volume claims from a role
-func getVolumeClaims(role *model.Role) []v1.PersistentVolumeClaim {
+func getVolumeClaims(role *model.Role, createHealmChart bool) []v1.PersistentVolumeClaim {
```
mook-as: Not a fan of bool arguments (hard to figure out what they are).

Random thought:
```golang
type KubeGeneratorType bool
const (
	KubeGeneratorRaw false
	KubeGeneratorHelm true
)
```

… okay that probably doesn't fit in this PR does it. Let's leave it alone for now and we can change later.

app/fissile.go
```diff
@@ -951,6 +951,14 @@ func (f *Fissile) generateHelmValues(outputDir string, rolesManifest *model.Role
 		_, err = outputFile.Write(buf)
 	}
 	if err == nil {
+		_, err = outputFile.Write([]byte(`kube:
```
mook-as: Not a fan of embedding YAML in the source like this. (I'm less against doing it in tests, but this isn't that.)

Is there something we can do here better? Put it in `buf` next to `env`?

kube/stateful_set.go
jandubois: I agree with the comment and will change it if there is any other change required.

app/fissile.go
jandubois: Please suggest what you would like to see.  I got these from Aaron (and they don't even cover the `kube.external_ip`:

```
	buf, err := yaml.Marshal(map[string]interface{}{
		"env": env,
		"k8s": map[string]interface{}{"storage-class": map[string]interface{}{"shared": "shared", "persistent": "persistent"}}},
	)
```

and

```
type A struct {
   Env string `yaml:"env"`
   K8s B `yaml:"k8s"`
}

type B struct {
   StorageClass StorageClass `yaml:"storage-class"`
}

type StorageClass struct {
   Shared string `yaml:"shared"`
   Persistent string `yaml:"persistent"`
}

yaml.Marshal(A{ 
  Env: env,
  K8s: B{ StorageClass{ Shared: "shared", Persistent: "persistent" } },
})
```

I consider neither of them an improvement over the literal YAML.

mook-as: I was thinking the first option, mainly due to indenting issues with the literal YAML.  (But without the `k8s` thing all on one line, probably…)

If you feel strongly about it, though, we can go with what you have.

kube/stateful_set.go
jandubois: An even worse example is `headless bool, public bool`: https://github.com/SUSE/fissile/blob/master/kube/service.go#L42

app/fissile.go
jandubois: I'm unable to come up with any way that is remotely readable and also acceptable to the Go compiler.  Feel free to attach exact code (that has passed the formatter and compiler).

I think Go is simply not a suitable language for dynamic data structures like this.

mook-as: It is possible that my readability bar is set too low due to years to terrible language abuse:
```golang
	buf, err := yaml.Marshal(map[string]interface{}{
		"env": env,
		"kube": map[string]interface{}{
			"external_ip": "192.168.77.77",
			"storage_class": map[string]interface{}{
				"persistent": "persistent",
				"shared":     "shared",
			},
		},
	})
	if err == nil {
		_, err = outputFile.Write(buf)
	}
```

jandubois: That's fine; I'll take. At least it is better than what I was able to get.

I still think the literal YAML is much easier to read, but hey, at least this version is type-safe (or is it?). 😄

jandubois: Fixed in c58dcd86b623cf1991c3d9ab9599f5fdf00500b3

kube/stateful_set.go
jandubois: > Let's leave it alone for now and we can change later.

I've changed my mind on this and will now go with the part of your feedback quoted above.

PR #262, opened 2017-07-17
Add `stop-on-failure` label

alexjh:
A `bosh-task` with the tag `stop-on-failure` is a basic Pod that won't
restart, even if it fails. This can be used to run tests where you need
it to fail, and not restart the pod, so the logs can be examined.

A job always tries to run to successful completion:
https://github.com/kubernetes/kubernetes/issues/20255

### DIFF
--- a/kube/pod_test.go
+++ b/kube/pod_test.go
@@ -8,12 +8,11 @@ import (
 	"strings"
 	"testing"
 
-	yaml "gopkg.in/yaml.v2"
-
 	"github.com/SUSE/fissile/model"
 	"github.com/SUSE/fissile/testhelpers"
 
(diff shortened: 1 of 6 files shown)


kube/pod_test.go
```diff
@@ -1,19 +1,25 @@
...
 
 import (
+	"bytes"
 	"encoding/base64"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
+	yaml "gopkg.in/yaml.v2"
```
mook-as: This should probably go live with `assert` and other packages at the end of the `import` section

alexjh: Fixed in 11c4ec9

PR #266, opened 2017-07-28
model/roles: Use relative paths in script signatures

mook-as:
This ensures that two builds using different checkouts can still get the same signature as long as they are at the same commit / using the same inputs.  This is necessary for CI as that puts each build in a checkout
with a random path.

### DIFF
--- a/model/roles.go
+++ b/model/roles.go
@@ -400,13 +400,11 @@ func (r *Role) GetScriptPaths() map[string]string {
 func (r *Role) GetScriptSignatures() (string, error) {
 	hasher := sha1.New()
 
-	i := 0
 	paths := r.GetScriptPaths()
-	scripts := make([]string, len(paths))
+	scripts := make([]string, 0, len(paths))
 
-	for _, f := range paths {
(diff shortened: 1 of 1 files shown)


model/roles.go
```diff
@@ -400,30 +400,28 @@ func (r *Role) GetScriptPaths() map[string]string {
...
 
-		f, err := os.Open(filename)
+		f, err := os.Open(paths[filename])
 		if err != nil {
 			return "", err
 		}
 
-		if _, err := io.Copy(hasher, f); err != nil {
+		_, err = io.Copy(hasher, f)
+		f.Close()
```
andreas-kupries: Could we use `defer f.Close()` to auto-close (would have to go to just after the os.Open).
Not sure how much this stacks(sic) things up. (Given the loop).

mook-as: We could, but (exactly due to the looping) I thought it would be better to avoid doing it.  I actually wrote that first 😄

andreas-kupries: Ok. No objections then.

PR #267, opened 2017-07-31
Add additional logging level

mook-as:
This is more for discussion than actual pull.

It would be nice to have something like an additional logging level, so that we can troubleshoot why image tags are changing.  I don't really like how invasive this patch is though; maybe something like a `termui.UI.Printf()` variant that took logging levels? Or just a levelled logger on top of a writer?


mook-as:
Stale PR; now that we have graphing (#314) we can see how that goes instead.  Closing PR.

PR #269, opened 2017-08-02
Add a release target and improve versioning

greygoo:
- Add a release target packaging the fissile sources
- Retrieve version information from a .version file if it exists

### DIFF
--- a/make/release
+++ b/make/release
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
+SRC=`basename $(pwd)`
+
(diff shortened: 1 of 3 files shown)


make/release
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
+SRC=`basename $(pwd)`
```
jimmykarily: Not sure if its an overkill but there are ["safer" ways to get the script's dir](https://stackoverflow.com/questions/59895/getting-the-source-directory-of-a-bash-script-from-within). It it is working maybe we can get away with this.

PR #271, opened 2017-08-03
Add docker-network-mode argument to fissile build packages

jimmykarily:
or any other option here:
https://docs.docker.com/engine/userguide/networking/

it delegates to NetworkMode in this project:
https://github.com/fsouza/go-dockerclient/blob/4b8b44fdd6d7dc8ef1acec5f21582f339cfbdaaa/container.go#L697

Builds on OBS will work if `--network=host` otherwise we get errors (e.g. can't resolve "localhost" to "127.0.0.1").

@viovanov

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -299,7 +299,7 @@ func newPropertyInfo(maybeHash bool) *propertyInfo {
 }
 
 // Compile will compile a list of dev BOSH releases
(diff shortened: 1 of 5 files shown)


jimmykarily:
Maybe the name of the env var should be FISSILE_DOCKER_NETWORK_MODE. Also, I'm not sure if it should be a command line option instead.

jimmykarily:
Changed to command line argument through cobra/viber

PR #281, opened 2017-09-21
Make sure all env vars in kube config files are quoted strings

jandubois:


helm/config.go
```diff
@@ -183,22 +183,7 @@ func (scalar *Scalar) Value() string {
...
-	// Calculate proper indentation for data lines.
-	if strings.HasSuffix(prefix, ":") {
-		prefix = strings.Repeat(" ", strings.LastIndex(prefix, " ")+1+enc.indent)
-	} else {
-		prefix = strings.Repeat(" ", len(prefix)+1)
-	}
-	for _, line := range strings.Split(scalar.value, "\n") {
-		fmt.Fprintln(enc, prefix+line)
-	}
+	fmt.Fprintln(enc, prefix+" "+strings.Replace(scalar.value, "\n", "\\n", -1))
```
mook-as: Please test other funny values, like `\r` and `\t`?

helm/config_test.go
```diff
@@ -645,20 +645,14 @@ Mapping:
...
+	root := NewNodeMapping("Scalar", NewScalar("\"foo\nbar\nbaz\""))
+	list := NewNodeList(NewScalar("\"one\ntwo\nthree\""))
 	root.AddNode("List", list)
 
 	expect := `---
-Scalar: |-
-    foo
-    bar
-    baz
+Scalar: "foo\nbar\nbaz"
```
mook-as: Ah, because the `equal()` function checks string equality, not YAML-serialization equality.

PR #282, opened 2017-09-26
Implement affinity declarations

jandubois:


helm/config.go
```diff
@@ -179,6 +217,10 @@ func (scalar *Scalar) String() string {
 
 // Value returns the value of a scalar node.
```
jandubois: The comment should mention that this method returns string values *without* the surrounding quotes.

kube/job.go
```diff
@@ -17,16 +17,21 @@ func NewJob(role *model.Role, settings *ExportSettings) (helm.Node, error) {
...
-		podTemplate.Get("spec").Get("restartPolicy").SetValue("OnFailure")
+		podTemplate.Get("spec/restartPolicy").SetValue("OnFailure")
 	default:
 		return nil, fmt.Errorf("Role %s has unexpected flight stage %s", role.Name, role.Run.FlightStage)
 	}
 
-	job := newTypeMeta("extensions/v1beta1", "Job")
-	job.AddNode("metadata", helm.NewMapping("name", role.Name))
-	job.AddNode("spec", helm.NewNodeMapping("template", podTemplate))
+	kind := "extensions/v1beta1"
```
jandubois: `kind` should be `api` or `apiVersion`

kube/deployment.go
```diff
@@ -19,20 +20,48 @@ func NewDeployment(role *model.Role, settings *ExportSettings) (helm.Node, helm.
...
-	spec.AddNode("selector", newSelector(role.Name))
-	spec.AddNode("template", podTemplate)
+	spec.Add("selector", newSelector(role.Name))
+	spec.Add("template", podTemplate)
 
 	deployment := newKubeConfig("extensions/v1beta1", "Deployment", role.Name, helm.Comment(role.GetLongDescription()))
-	replicaCheck(role, deployment, spec, svc, settings)
-	deployment.AddNode("spec", spec.Sort())
+	err = replicaCheck(role, deployment, spec, svc, settings)
+	deployment.Add("spec", spec.Sort())
```
jandubois: Maybe `spec` should be added to `deployment` first and not passed as an explicit parameter to `replicaCheck`. The function already locates other fields inside `spec` via `Get()` calls, so why not locate `spec` itself too and get rid of the redundant parameter?

helm/config.go
```diff
@@ -153,19 +154,56 @@ type Node interface {
 	write(enc *Encoder, prefix string)
 }
 
+// NewNode creates a new scalar node.
+func NewNode(value interface{}, modifiers ...NodeModifier) Node {
+	var node Node
+	if value == nil {
+		node = &Scalar{value: "~"}
+	} else if _, ok := value.(Node); ok {
+		node = value.(Node)
```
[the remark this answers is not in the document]
jandubois: Not without declaring `ok` at the top. But then I find the current code clearer.

helm/config.go
```diff
@@ -153,19 +154,56 @@ type Node interface {
...
+	} else {
+		switch reflect.TypeOf(value).Kind() {
+		case reflect.Map:
+			valueOf := reflect.ValueOf(value)
+			mapping := NewEmptyMapping()
+			for _, key := range valueOf.MapKeys() {
+				mapping.Add(key.Interface().(string), valueOf.MapIndex(key).Interface())
```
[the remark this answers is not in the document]
jandubois: Yes, it will fail if the key in the YAML is not a string.  That shouldn't happen if the definition in the role manifest itself is valid, as I don't think kube uses non-string keys.

But what should we do? Catch the type mis-match and call panic ourselves? That doesn't seem better than what we get by default.

kube/pod.go
```diff
@@ -59,24 +59,24 @@ func NewPodTemplate(role *model.Role, settings *ExportSettings) (helm.Node, erro
 	container := helm.NewEmptyMapping()
```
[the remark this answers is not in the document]
jandubois: Yeah, `NewEmptyMapping` and `NewEmptyList` are only needed anymore if you want to set a modifier on the node, which we seem to do only in one place outside the tests:

```
$ ag 'NewEmpty(Mapping|List)\((?!\))' kube/
kube/values.go
39:		entry := helm.NewEmptyMapping(helm.Comment(role.GetLongDescription()))
```

So I'm going to remove the `NewEmpty*` variants...

helm/config.go
```diff
@@ -266,58 +296,36 @@ func NewEmptyMapping(modifiers ...NodeModifier) *Mapping {
...
+// Add a single named node at the end of the list.
+func (mapping *Mapping) Add(name string, value interface{}, modifiers ...NodeModifier) {
+	mapping.nodes = append(mapping.nodes, namedNode{name: name, node: NewNode(value, modifiers...)})
 }
 
-// Get returns the named node, or nil if the name cannot be found.
+// Get returns the named node, or nil if the name cannot be found. Name may be
+// a "/" separated "path" of nested names, which will be treated as a string of
+// Get() calls, but with the advantage of not crashing if any intermediate node
+// does not exist.
```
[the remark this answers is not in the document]
jandubois: There normally shouldn't be slashes. Except when kube uses its weird keys for namespacing, like `scheduler.alpha.kubernetes.io/affinity`. So maybe `Get(name ...string)` is better, even though slightly uglier to read.

helm/config.go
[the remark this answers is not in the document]
mook-as: I was hoping we'd be able to propagate the error up with enough information for the user to figure out where they messed things up and fix it.  But that was before I realized that this function had no error returning capability…

helm/config.go
```diff
@@ -153,19 +154,56 @@ type Node interface {
...
 
-// NewScalar creates a scalar node and initializes shared fields.
-func NewScalar(value string, modifiers ...NodeModifier) *Scalar {
-	scalar := &Scalar{value: value}
-	scalar.Set(modifiers...)
-	return scalar
-}
-
 // SetValue updates the value of a scalar node.
 func (scalar *Scalar) SetValue(value string) {
```
jandubois: `SetValue()` should take an `interface{}` argument instead of a `string`, so the user doesn't have to deal with quoting.

helm/config.go
jandubois: https://github.com/SUSE/fissile/pull/282/commits/0ed98d295f2ada6bfa5590e0688a379f9f61d580

kube/job.go
jandubois: https://github.com/SUSE/fissile/pull/282/commits/13b9e729e2083fd62c79884643da80aba2183073

kube/deployment.go
jandubois: https://github.com/SUSE/fissile/pull/282/commits/27d37c3de22062e27a141fd59731071c3561ee30

helm/config.go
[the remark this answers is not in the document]
jandubois: https://github.com/SUSE/fissile/pull/282/commits/70046690733282beb807077cd2ff930bea1b5dbf

kube/job.go
```diff
@@ -17,16 +17,21 @@ func NewJob(role *model.Role, settings *ExportSettings) (helm.Node, error) {
...
+	if settings.CreateHelmChart {
+		kind = "{{ if and (eq (int .Capabilities.KubeVersion.Major) 1) (le (int .Capabilities.KubeVersion.Minor) 5) }}"
```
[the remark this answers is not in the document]
jandubois: https://github.com/SUSE/fissile/pull/282/commits/8f4a7e4801eb55abf6ae4fdc8b27290208c247c8

kube/pod.go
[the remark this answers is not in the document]
jandubois: https://github.com/SUSE/fissile/pull/282/commits/f9959d2dd8d77027252bcfb55998671007208502

helm/config.go
jandubois: https://github.com/SUSE/fissile/pull/282/commits/0ed98d295f2ada6bfa5590e0688a379f9f61d580

helm/config.go
```diff
@@ -153,32 +149,69 @@ type Node interface {
...
+		node = &Scalar{value: "~"}
+	} else if _, ok := value.(Node); ok {
+		node = value.(Node)
+	} else {
+		switch reflect.TypeOf(value).Kind() {
+		case reflect.Map:
+			valueOf := reflect.ValueOf(value)
+			mapping := NewMapping()
+			for _, key := range valueOf.MapKeys() {
+				mapping.Add(key.Interface().(string), valueOf.MapIndex(key).Interface())
```
mook-as: I think I still want this to be `mapping.Add(fmt.Sprintf("%s", key.Interface()), …)`. Thoughts?
That won't save you from bad input, but it _will_ save you from a panic. That way the kube validation (on the server end) will deal with the problem.

PR #283, opened 2017-09-26
Include extra information for BOSH links

mook-as:
This PR is in support of using BOSH 2.0 features.  Here, it adds support for resolving links (at build time, of course) and putting that information into the image so later configgin changes can pick it up.


model/job.go
```diff
@@ -17,21 +17,46 @@ import (
...
+}
+
+// JobLinkProvides describes the BOSH links a job provides
+type JobLinkProvides struct {
+	Name       string
+	Role       *Role
+	Job        *Job
+	Type       string   `json:"type"`
+	Properties []string `json:"properties"`
+	providers  []jobLinkProvider
```
jandubois: Why is providers a list? It looks like it is either empty or just one element. Since the members of the type are pointers, you could just use `nil` to indicate that the list is empty?

Also, how is `providers.Role` different from `Role` (and `providers.Job different from `Job`)?

model/job.go
```diff
@@ -240,20 +287,45 @@ func (j *Job) MergeSpec(otherJob *Job) {
 	j.Properties = append(j.Properties, otherJob.Properties...)
 }
 
+type configConsumes struct {
```
jandubois: This is effectively the same type as `jobLinkProvider`.  Maybe there is a better name that describes this type that covers both uses? What about `jobLink`, or `linkTarget`?

model/job.go
```diff
@@ -230,20 +274,43 @@ func (j *Job) MergeSpec(otherJob *Job) {
...
-	config, err := initializeConfigJSON()
-	if err != nil {
-		return nil, err
-	}
-
-	config["job"].(map[string]interface{})["name"] = role.Name
+	var config struct {
+		Job struct {
+			Name      string `json:"name"`
+			Templates []struct {
```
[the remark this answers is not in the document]
mook-as: Oops, missed this comment. I'll add something for this.
This isn't a `JobTemplate` because its' just being used to serialize to JSON… though I should think about it.

mook-as:
This is wrong — I've simplified too much, and I'm using the name of the consumer-side job for the provider information. (This leads to loggergator_trafficcontroller in the loggregator role wanting to look for itself in the doppler role).

model/job.go
```diff
@@ -263,7 +326,16 @@ func (j *Job) WriteConfigs(role *Role, lightOpinionsPath, darkOpinionsPath strin
 	if err != nil {
 		return nil, err
 	}
-	config["properties"] = properties
+	config.Properties = properties
+
+	for _, provider := range j.LinkProvides {
+		config.ExportedProperties = append(config.ExportedProperties, provider.Properties...)
```
jandubois: Is it ok if there a multiple entries for the same property in this list? If not, wouldn't this be better as a map?

model/job.go
```diff
@@ -263,7 +326,16 @@ func (j *Job) WriteConfigs(role *Role, lightOpinionsPath, darkOpinionsPath strin
...
+	}
+	for _, consumeTarget := range j.LinkConsumes {
+		if consumeTarget.Role != "" && consumeTarget.Job != "" {
+			config.Consumes[consumeTarget.Name] = consumeTarget
```
jandubois: What happens when `consumeTarget.Name` is `""`?

model/job.go
mook-as: I don't think it would actually _hurt_.

model/job.go
mook-as: That would mean the provider had no name too, right? (Because otherwise `.Role` and `.Job` wouldn't have been filled in)

I guess we just have a very silly entry… I can skip that case, sure.

jandubois: Yes, the provider must have a name. But do you copy the name into the consumer when you resolve nameless links? (Just asking because it will be faster for you to check, than for me)

mook-as: Yes, we do (and there's tests for it).

model/roles.go
```diff
@@ -380,9 +381,61 @@ func (m *RoleManifest) GetRoleManifestDevPackageVersion(roles Roles, opinions *O
...
+	// Resolve the consumers
+	for _, role := range m.Roles {
+		// Prefer finding providers in the same role
+		providingRoles := append(Roles{role}, m.Roles...)
+		for _, job := range role.Jobs {
+		iterateOverConsumers:
+			for idx, consumer := range job.LinkConsumes {
+				for _, providingRole := range providingRoles {
+					for _, providingJob := range providingRole.Jobs {
+						for _, provider := range providingJob.LinkProvides {
```
mook-as: Not enjoying the sextuply-nested loop, but this _does_ seem like cleaner code and probably doesn't matter…

model/roles.go
```diff
@@ -380,9 +381,61 @@ func (m *RoleManifest) GetRoleManifestDevPackageVersion(roles Roles, opinions *O
 
 // LookupRole will find the given role in the role manifest
 func (m *RoleManifest) LookupRole(roleName string) *Role {
+	if m.rolesByName == nil {
```
jandubois: I would drop the `rolesByName` cache and do a search for every lookup. I don't think you'll ever even have a 100 roles in a single manifest, so this feels like premature optimization.

If you want to keep the cache, then maybe extend the condition to be robust against manifest changes after the first lookup:

```
if m.RolesByName == nil || m.RolesByName[role.Name] == nil {
```

That of course assumes that most lookups will be successful. Because you'll rebuild the cache on each failed lookup. But I would just drop the cache. 😃

model/roles.go
jandubois: Definitely cleaner than the previous version. I like the simplicity of the logic.

model/roles_test.go
```diff
@@ -526,3 +527,176 @@ func TestLoadRoleManifestRunGeneral(t *testing.T) {
...
+
+	roleManifestPath := filepath.Join(workDir, "../test-assets/role-manifests/multiple-good.yml")
+	roleManifest, err := LoadRoleManifest(roleManifestPath, releases)
+	assert.NoError(t, err)
+
+	errList := roleManifest.resolveLinks()
+	assert.Empty(t, errList)
+
+	role := roleManifest.LookupRole("myrole")
+	var job *Job
```
jandubois: Why not

```
job := role.LookupJob("ntpd")
```

model/job.go
```diff
@@ -17,21 +17,38 @@ import (
...
+	Properties []string
+}
+
+// JobLinkConsumes describes the BOSH links a job consumes
+type JobLinkConsumes struct {
+	Name     string `json:"-"`
+	Type     string `json:"-"`
+	Optional bool   `json:"-"`
+	Role     string `json:"role"` // Name of role in which the resolved job is running
+	Job      string `json:"job"`  // Name of job that this resolves to
```
jandubois: I wonder if these 2 fields should be `RoleName` and `JobName` instead? Because right now I would expect them to be of type `*Role` and `*Job` instead of `string`, based on their name.

mook-as: e57c5c1aa5a881789004d3a7cde234396d34efe3

model/roles_test.go
mook-as: Because that didn't exist yet when I wrote that line 😊 
Fixed in c854927c6dc05962f2e2c02c4f7ca472eab163bf

PR #284, opened 2017-09-27
Use template for docker images

alexjh:


kube/pod.go
```diff
@@ -110,8 +110,11 @@ func getContainerImageName(role *model.Role, settings *ExportSettings) (string,
 	if err != nil {
 		return "", err
 	}
-	imageName := builder.GetRoleDevImageName(settings.Registry, settings.Organization, settings.Repository, role, devVersion)
-	return imageName, nil
+	imageName := builder.GetRoleDevImageName("", "", settings.Repository, role, devVersion)
+	value := "{{ default \"docker.io\" .Values.kube.registry -}}/"
```
jandubois: Why don't we put the `docker.io` default value into `values.yaml` instead?

I don't have strong feelings either way, but my gut instinct is to define a default value in one place, and not repeat it wherever it is used. But then we have the `require` mechanism, which can't be specified in `values.yaml`, so maybe it is better the way it is? Can't make up my mind. 😄

kube/pod.go
```diff
@@ -110,8 +110,11 @@ func getContainerImageName(role *model.Role, settings *ExportSettings) (string,
 	if err != nil {
 		return "", err
 	}
-	imageName := builder.GetRoleDevImageName(settings.Registry, settings.Organization, settings.Repository, role, devVersion)
-	return imageName, nil
+	imageName := builder.GetRoleDevImageName("", "", settings.Repository, role, devVersion)
```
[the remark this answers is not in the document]
alexjh: Aha, thank you! That fixes the tests as well.

kube/pod.go
```diff
@@ -110,11 +110,18 @@ func getContainerImageName(role *model.Role, settings *ExportSettings) (string,
...
-	value := "{{ default \"docker.io\" .Values.kube.registry -}}/"
-	value += "{{- .Values.kube.organization -}}/"
-	value += imageName
-	return value, nil
+
+	var imageName string
+	if settings.CreateHelmChart {
+		name := builder.GetRoleDevImageName("", "", settings.Repository, role, devVersion)
+		imageName = "{{ default \"docker.io\" .Values.kube.registry -}}/"
+		imageName += "{{- .Values.kube.organization -}}/"
```
jandubois: Not sure why you use `{{- ... -}}` instead of plain `{{ ... }}`. The dashes are eating up whitespace outside the template, but you don't have any whitespace there.

kube/pod.go
```diff
@@ -110,11 +110,18 @@ func getContainerImageName(role *model.Role, settings *ExportSettings) (string,
...
+		imageName += name
```
jandubois: I wonder if you could just write:

```
registry := `{{ ... .Values.kube.registry }}`
org := `{{ .Values.kube.organization }}`
imageName = builder.GetRoleDevImageName(registry, org, ...)
```

jandubois: Hmm, would require an exception clause in `SanitizeDockerName` to do nothing if the name starts with `{{` and ends with `}}`.

kube/pod.go
alexjh: Oh, I misread the docs. I thought it would trim whitespace from the variables being replaced in the template. You'd have to try pretty hard to break it in that way though.

kube/pod.go
alexjh: Actually, a default is still needed. :( Still thinking about that one.

jandubois: Why is a default needed, the code handles `""` correctly:

https://github.com/SUSE/fissile/blob/master/builder/role_image.go#L505-L507

kube/pod.go
jandubois: Yeah, to trim whitespace from the variable you would have to pass it through the `trim` template function: `{{ .Values.FOO | trim }}`

http://masterminds.github.io/sprig/strings.html

kube/pod.go
alexjh: Ok, updated the sanitize function.

The image name generated by the helm template is `docker.io/splatform/foo:1234`, and *fortunately* the docker client treats that the same as `splatform/foo:1234`, so no changes are needed to the SCF project. I was worried that it would force a change to the build scripts to include `docker.io` in the image names, but that doesn't seem to be the case.

alexjh:
Updated the PR to handle auth.

It adds two options to fissile: `docker-username` and `docker-password`. These get set in the `values.yaml` file, and are used in the new `registry-secrets.yaml` file.

If you're curious about the encoded string in the secrets file, you can examine it after it's been installed:
`kubectl -n uaa get secret registry-credentials -o json | jq -r '.data.".dockercfg"' | base64 -d | jq`

kube/registry_credentials.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"github.com/SUSE/fissile/helm"
+)
+
+// MakeRegistryCredentials generates a template that contains Docker Registry credentials
+func MakeRegistryCredentials(createHelmChart bool) (helm.Node, error) {
+
+	value := ""
+	if createHelmChart {
+		value = "{{ printf \"{%s:{%s:%s,%s:%s,%s:%s}}\" (.Values.kube.registry.hostname | quote) (printf \"%s\" \"username\" | quote) (default \"\" .Values.kube.registry.username | quote) ( printf \"%s\" \"password\" | quote) (default \"\" .Values.kube.registry.password | quote) ( printf \"%s\" \"auth\" | quote) ( printf \"%s:%s\" .Values.kube.registry.username .Values.kube.registry.password | b64enc | quote ) | b64enc }}"
```
jandubois: This line is too long; I would like to see it broken up into smaller pieces that are then put together. Or find some other way to line wrap at significant points.

Use Go's backquotes to avoid having to escape double-quotes inside the string: `` `printf "user" ...` ``.

Also isn't `(printf "%s" "username" | quote)` the same as `("username" | quote)` (or `(printf "%q" "username")`)?

kube/pod.go
```diff
@@ -68,8 +68,12 @@ func NewPodTemplate(role *model.Role, settings *ExportSettings) (helm.Node, erro
 	container.Add("readinessProbe", readinessProbe)
 	container.Sort()
 
+	imagePullSecrets := helm.NewMapping()
+	imagePullSecrets.Add("name", "registry-credentials")
```
jandubois: If a mapping has just a single entry then I would add that in the constructor call:

```
imagePullSecrets := helm.NewMapping("name", "registry-credentials")
```

kube/pod.go
```diff
@@ -68,8 +68,12 @@ func NewPodTemplate(role *model.Role, settings *ExportSettings) (helm.Node, erro
...
+
 	spec := helm.NewMapping()
 	spec.Add("containers", helm.NewList(container))
+	spec.Add("imagePullSecrets", helm.NewList(imagePullSecrets))
```
jandubois: I have not actually tested this change yet, but I'm wondering if specifying "empty" secrets work, or if trying to log in with an empty username/password in a registry that doesn't support auth will throw an error.

kube/registry_credentials.go
```diff
@@ -0,0 +1,23 @@
...
+	}
+
+	data := helm.NewMapping()
+	data.Add(".dockercfg", helm.NewNode(value))
```
jandubois: This can just be

```
data := helm.NewMapping(".dockercfg", value)
```

app/fissile.go
```diff
@@ -853,15 +853,49 @@ func (f *Fissile) GenerateKube(roleManifestPath, outputDir, repository, registry
...
-		if err = f.generateHelmValues(outputDir, roleManifest, defaults); err != nil {
+		if err = f.generateHelmValues(outputDir, roleManifest, defaults, registry, username, password, organization); err != nil {
 			return err
 		}
 	}
 
 	return f.generateKubeRoles(outputDir, repository, registry, organization, fissileVersion, roleManifest, defaults, refs, useMemoryLimits, createHelmChart, opinions)
 }
 
+func (f *Fissile) generateRegistrySecret(outputDir string, secret helm.Node, createHelmChart bool) error {
```
jandubois: We now have at least 3 functions that are all just writing out a `helm.Node` to a file, with very little variation, so maybe this can be refactored into a single function?

kube/registry_credentials.go
alexjh: Fixed in e7c8af4

kube/pod.go
alexjh: I considered putting a `Block` around this if the username is null in `values.yaml`, it might be necessary. I'll try setting up a local registry to test this.

app/fissile.go
alexjh: Fixed in 525c1bd

kube/registry_credentials.go
alexjh: What do you think of 86e4600?

jandubois:
> What do you think of 86e4600?

Yeah, that is much better. I guess I could still criticize that `(print ... | b64end )` should have the space before the right parens stripped for balance, but I think I'll let that slide. 😄

jandubois:
Replaced by #289.

PR #285, opened 2017-10-02
Helm chart improvements

jandubois:
Add role description to Job objects.

Sort replica min/max to the top of Deployment spec files.

### DIFF
--- a/kube/stateful_set.go
+++ b/kube/stateful_set.go
@@ -36,7 +36,7 @@ func NewStatefulSet(role *model.Role, settings *ExportSettings) (helm.Node, helm
 	statefulSet.Add("spec", spec)
 	err = replicaCheck(role, statefulSet, svcList, settings)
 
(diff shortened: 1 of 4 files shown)


kube/stateful_set.go
```diff
@@ -36,7 +36,7 @@ func NewStatefulSet(role *model.Role, settings *ExportSettings) (helm.Node, helm
 	statefulSet.Add("spec", spec)
 	err = replicaCheck(role, statefulSet, svcList, settings)
 
-	return statefulSet.Sort(), svcList, err
+	return statefulSet, svcList, err
```
alexjh: Why isn't this being sorted any more? It's not clear to me why this is being done, I thought the problem was with deployments only.

alexjh: Oh, nevermind, I see now that it was moved to the common `replicaCheck` function.

PR #287, opened 2017-10-06
Add extra cli arguments to add more information for image tag calculation.

andreas-kupries:
https://trello.com/c/HaqnE735/391-fissile-doesnt-have-enough-information-to-generate-good-docker-image-tags-when-scf-changes

Implementation should be complete. Unit tests pass.
Still needs testing within the context of a vagrant box building images.

Hm ... Coming to dislike the current name of the new option (`--version`).
The env var for that is `FISSILE_VERSION`, and completely does __not__ mean that.
To confusing.
A better env name would be `FISSILE_TAG_EXTRA`, and option `--tag-extra`.
Going for that now.

SCF PR = https://github.com/SUSE/scf/pull/1148 (🚧 No fissile master build yet)

This PR now ready for review.

### DIFF
--- a/cmd/build-helm.go
+++ b/cmd/build-helm.go
@@ -11,7 +11,7 @@ var (
 	flagBuildHelmOutputDir       string
 	flagBuildHelmDefaultEnvFiles []string
 	flagBuildHelmUseMemoryLimits bool
-	flagBuildHelmVersion         string
+	flagBuildHelmTagExtra        string
 )
 
 // buildHelmCmd represents the helm command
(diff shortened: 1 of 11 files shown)


cmd/build-helm.go
```diff
@@ -84,5 +86,12 @@ func init() {
 		"Include memory limits when generating helm chart",
 	)
 
+	buildHelmCmd.PersistentFlags().StringP(
+		"version",
```
mook-as: Note that we already have a [`FISSILE_VERSION`](https://github.com/SUSE/scf/blob/10b4a5ccb4d09871923bcc60c6274b692ca1a589/bin/common/versions.sh#L12) variable, which will get picked up because of viper/cobra.  You might need a different name, or change that.

andreas-kupries: x-ed, see my latest musing in main comment. Changing to `--tag-extra` now.

PR #288, opened 2017-10-11
model: same job of different roles should resolve links differently

mook-as:
When a job is used in multiple roles, we might need to resolve its BOSH
link consumers differently.  For example, many roles have a consul_agent
job; when resolving links to consul, they should find it on the same
role instead of looking in a different role.


mook-as:
Found a bug with the generated link info; uaa-mysql won't come up.

model/job.go
```diff
@@ -17,36 +17,43 @@ import (
...
+	Templates     []*JobTemplate
+	Packages      Packages
+	Path          string
+	Fingerprint   string
+	SHA1          string
+	Properties    []*JobProperty
+	Version       string
+	Release       *Release
+	LinkProviders map[string]jobProvidesInfo
+	LinkConsumers []jobConsumesInfo
```
jandubois: Not sure why `LinkProviders` is a map and `LinkConsumers` just a plain array.

I guess there is currently no need to look up a consumer by name, but it would feel better to me if both were stored the same way for symmetry reasons.

jandubois: It would also more closely match the `RoleJob` struct in `model/roles.go`.

In that struct these 2 fields are called `Provides` and `Consumes` instead of `LinkProviders` and `LinkConsumers` (and both are maps). Maybe the naming should be unified?

model/roles.go
```diff
@@ -338,18 +340,28 @@ func LoadRoleManifest(manifestFilePath string, releases []*Release) (*RoleManife
 				continue
 			}
 
-			role.Jobs = append(role.Jobs, job)
+			roleJob.Job = job
+
+			for name, info := range roleJob.Consumes {
+				info.Name = name
+				roleJob.Consumes[name] = info
```
jandubois: Isn't this already set? Why else would the `range` operator return this combination of `name` and `info`?

model/roles.go
```diff
@@ -373,38 +385,117 @@ func (m *RoleManifest) LookupRole(roleName string) *Role {
 func (m *RoleManifest) resolveLinks() validation.ErrorList {
 	errors := make(validation.ErrorList, 0)
 
-	// Resolve the consumers
+	// Build mappings of providers by name, and by type.  Note that the names
+	// involved here are the aliasesp, where appropriate.
```
jandubois: "aliasesp"?

model/roles.go
mook-as: Because this is loaded from user-provided role-manifest, where it's probably
```yaml
…
consumes:
    thing:
        # No `name` subkey here
        type: mysql
```

jandubois: Ah, I see now; it is the whole `range` returning values tripping me up again. ☹️ 

So to avoid copying structs back and forth it could be

```
for name := range roleJob.Consumes {
    roleJob.Consumes[name].Name = name
}
```

mook-as: Yes, except that as far as I can tell that's not valid in golang for reasons I don't understand. It just throws compiler errors at me.

model/roles.go
mook-as: It's a very lispy alias? (Fixed)

model/job.go
mook-as: I'll work on unifying the name; the issue here is that consumer info might not have names (because they're matching by types), so there are no unique keys to use here. I'd have to switch to using lists for everything instead.

mook-as: Hmm. Since `RoleJob` embeds `Job`, it's actually accessible as `roleJob.LinkProviders` right now.  That's… going to be totally confusing.  Naming suggestions?

jandubois: I think using `Provides` and `Consumer` for both should be fine. When you are dealing with a `RoleJob` you really want the Union of both providers or consumers of the `RoleJob` and the referenced `Job`, right?

I kind of view the `Job` as the base class, with the `RoleJob` providing additional specialization.

Or maybe I'm already totally confused and am not making much sense. There is a reason I'm so slow in reviewing this PR...

mook-as: Renamed in 6a537f1

model/roles.go
```diff
@@ -373,38 +385,117 @@ func (m *RoleManifest) LookupRole(roleName string) *Role {
...
-							consumer.RoleName = providingRole.Name
-							consumer.JobName = providingJob.Name
-							continue iterateOverConsumers
-						}
-					}
+		for _, roleJob := range role.RoleJobs {
+			var availableProviders []string
+			for availableName, availableProvider := range roleJob.Job.AvailableProviders {
+				availableProviders = append(availableProviders, availableName)
+				providersByType[availableProvider.Type] = append(providersByType[availableProvider.Type], jobProvidesInfo{
```
jandubois: Throw an error if `availableProvider.Type` is the empty string?

model/roles.go
```diff
@@ -373,38 +385,117 @@ func (m *RoleManifest) LookupRole(roleName string) *Role {
...
 
-				// No valid providers
-				if !consumer.Optional {
+	// Resolve the consumers
+	for _, role := range m.Roles {
+		for _, roleJob := range role.RoleJobs {
+			expectedConsumers := make([]jobConsumesInfo, len(roleJob.Job.DesiredConsumers))
+			copy(expectedConsumers, roleJob.Job.DesiredConsumers)
+			if roleJob.ResolvedConsumers == nil {
+				roleJob.ResolvedConsumers = make(map[string]jobConsumesInfo)
```
jandubois: Why create an empty map? It is not getting filled in afaict.

And `range` can iterate over it just fine (it already does so in `LoadRoleManifest`).

model/job.go
```diff
@@ -295,21 +307,24 @@ func (j *Job) WriteConfigs(role *Role, lightOpinionsPath, darkOpinionsPath strin
...
 	config.ExportedProperties = make([]string, 0)
-	config.Consumes = make(map[string]*JobLinkConsumes, 0)
+	config.Consumes = make(map[string]jobLinkInfo)
 
 	config.Job.Name = role.Name
 
-	for _, roleJob := range role.Jobs {
+	for _, roleJob := range role.RoleJobs {
 		config.Job.Templates = append(config.Job.Templates, jobConfigTemplate{roleJob.Name})
+		for _, consumer := range roleJob.ResolvedConsumers {
```
jandubois: Why are the consumers for *all* jobs of this role added here? It looks like the output should only describe this particular job (`j`) and not all of them.

model/job.go
```diff
@@ -295,21 +307,24 @@ func (j *Job) WriteConfigs(role *Role, lightOpinionsPath, darkOpinionsPath strin
...
 	config.Networks.Default = make(map[string]string)
 	config.ExportedProperties = make([]string, 0)
-	config.Consumes = make(map[string]*JobLinkConsumes, 0)
+	config.Consumes = make(map[string]jobLinkInfo)
 
 	config.Job.Name = role.Name
 
-	for _, roleJob := range role.Jobs {
+	for _, roleJob := range role.RoleJobs {
 		config.Job.Templates = append(config.Job.Templates, jobConfigTemplate{roleJob.Name})
```
jandubois: I know this hasn't been changed, but why are we merging all the templates from all the jobs inside the role containing `j`? Just trying to understand what it does.

mook-as: The templates are, as far aw I know, completely useless. Nobody actually looks at it.

Since this is `Job.WriteConfigs()`, we don't know which roleJob we're touching, though. So we still need to get the resolved consumers from there… I guess I can filter by checking if `roleJob.Job` is `j`?

model/job.go
mook-as: Oops, see previous reply. We have a `Job j`, which doesn't have resolved consumers (that's on `RoleJob`). We could filter to only the `RoleJob`s that have a `Job` of `j`?

jandubois: Why not turn this into a method on a `RoleJob`? The calling site already calls it that way:

https://github.com/SUSE/fissile/blob/marky-links-multiple-providers/builder/role_image.go#L163

It just turns into a `Job` method because the `RoleJob` has an unnamed `*Job` member that obviously has already been filled in.

model/roles.go
mook-as: It does get filled in the `range expectedConsumers` loop later in this function. But there we get the value (probably the zero value, if the consumer wasn't explicitly specified in the role manifest), which mean we should ensure the map isn't nil (but an empty map) before that point.

Yes, `range` can iterate over nil maps/slices. But indexing a nil map panics (whereas indexing an empty map just gives the zero value).

jandubois: What I see in that loop is just

```
					info := roleJob.ResolvedConsumers[name]
					info.Name = provider.Name
```

Since `ResolvedConsumers` is the empty map, `info` will be `nil`, so the assignment to `info.Name` will panic. If it never panics, it means we never get to that spot. Or I still don't understand it.

jandubois: Ah, nm, I missed the reference to the `zero` value, and that we get a value, not a pointer...

mook-as: Yep, if we don't create the map, then `info := roleJob.ResolvedConsumers[name]` will panic (since it involves dereferencing a nil pointer). But once that map exists (but is empty), the result is just an empty struct.

jandubois: I think I would still like to see the initialization either at the place where the whole struct is loaded (`LoadRoleManifest`?), or directly before the place where it _must_ be set to avoid the panic. Right now it is in a spot where it looks like it isn't really useful.

model/roles.go
```diff
@@ -373,38 +385,117 @@ func (m *RoleManifest) LookupRole(roleName string) *Role {
...
+						fmt.Sprintf(`consumer %s not found`, consumerAlias)))
+					continue
+				}
+				roleJob.ResolvedConsumers[consumerName] = jobConsumesInfo{
+					jobLinkInfo: provider.jobLinkInfo,
+				}
+
+				for i := len(expectedConsumers) - 1; i >= 0; i-- {
+					if expectedConsumers[i].Name == consumerName {
+						expectedConsumers = append(expectedConsumers[:i], expectedConsumers[i+1:]...)
```
jandubois: How can the same `consumerName` appear multiple times?

If it can't, why not add a `break` (and don't walk backwards because it doesn't make any difference then).

model/roles.go
jandubois: Actually, this should be done in `loadJobSpec`, both for `AvailableProviders` and `DesiredConsumers`.

mook-as: Hmm. Now I actually wonder if they _need_ to have types... I guess we'll enforce it for now and see what blows up.

model/roles.go
```diff
@@ -746,6 +752,56 @@ func (r *Role) calculateRoleConfigurationTemplates() {
...
+		config.ExportedProperties = append(config.ExportedProperties, provider.Properties...)
+	}
+
+	// Write out the configuration
+	jobJSON, err := json.MarshalIndent(config, "", "    ") // 4-space indent
+	if err != nil {
+		return nil, err
+	}
+
+	return jobJSON, nil
```
jandubois: Is it really necessary to make sure `jobJSON`  is `nil` when `err` isn't?  Otherwise this could simply be:

```
	// Write out the configuration
	return json.MarshalIndent(config, "", "    ") // 4-space indent
```

mook-as: I guess I just wanted to be consistent. Happy to change it if you really want to.

PR #289, opened 2017-10-16
Use template for docker images

jandubois:
This is a rebased version of #284 which had merge conflicts with `master`.

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -849,7 +849,7 @@ func (f *Fissile) GenerateKube(roleManifestPath, outputDir, repository, registry
 		return err
 	}
 
-	err = f.generateSecrets(outputDir, secrets, roleManifest, createHelmChart)
+	err = f.generateSecrets(outputDir, "secrets.yaml", secrets, createHelmChart)
 	if err != nil {
 		return err
 	}
@@ -859,12 +859,19 @@ func (f *Fissile) GenerateKube(roleManifestPath, outputDir, repository, registry
 		return err
 	}
 
-	if err = f.generateRegistrySecret(outputDir, registryCredentials, createHelmChart); err != nil {
+	if err = f.generateSecrets(outputDir, "registry-secret.yaml", registryCredentials, createHelmChart); err != nil {
 		return err
 	}
 
 	if createHelmChart {
-		if err = f.generateHelmValues(outputDir, roleManifest, defaults, registry, username, password, organization); err != nil {
+		values, err := kube.MakeValues(roleManifest, defaults, registry, username, password, organization)
+		if err != nil {
+			return err
+		}
+
+		// Export the default values for variables
+		outputPath := filepath.Join(outputDir, "values.yaml")
+		if err = f.writeHelmNode(outputPath, values); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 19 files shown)


kube/values.go
```diff
@@ -9,15 +9,15 @@ import (
 )
 
 // MakeValues returns a Mapping with all default values for the Helm chart
-func MakeValues(roleManifest *model.RoleManifest, defaults map[string]string) (helm.Node, error) {
+func MakeValues(settings *ExportSettings) (helm.Node, error) {
```
mook-as: The way this is being used, it might as well be `func (settings *ExportSettings) MakeValues() (helm.Node, error)`…

kube/values.go
```diff
@@ -69,9 +69,20 @@ func MakeValues(roleManifest *model.RoleManifest, defaults map[string]string) (h
 		sizing.Add(makeVarName(role.Name), entry.Sort(), helm.Comment(role.GetLongDescription()))
 	}
 
+	registry := settings.Registry
+	if registry == "" {
```
mook-as: Why do we need this again? (Comment please, either in code or as a `Comment()` for the node)

util/name_sanitization.go
```diff
@@ -10,5 +11,9 @@ var (
 
 // SanitizeDockerName makes a string conform with the rules for Docker names
 func SanitizeDockerName(name string) string {
+	if strings.HasPrefix(name, "{{") && strings.HasSuffix(name, "}}") {
```
mook-as: Should add a comment here about this being helm templates

app/fissile.go
```diff
@@ -824,83 +824,81 @@ func compareHashes(v1Hash, v2Hash keyHash) *HashDiffs {
...
-	}
-	_ = outputFile.Close()
-	return err
+	return f.writeHelmNode(secretsDir, fileName, secrets)
 }
 
-func (f *Fissile) generateHelmValues(outputDir string, roleManifest *model.RoleManifest, defaults map[string]string) error {
-	// Export the default values for variables
-	outputPath := filepath.Join(outputDir, "values.yaml")
+func (f *Fissile) writeHelmNode(dirName, fileName string, node helm.Node) error {
```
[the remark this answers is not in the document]
jandubois: I'm not going to change this; `HelmNode` references the `helm.Node` parameter and doesn’t map to `fissile build helm` or `fissile build kube`. We can rename it when we rename the `helm` package.

kube/registry_credentials.go
```diff
@@ -0,0 +1,38 @@
...
+		// }
+		//
+		// where "auth" is a base64 encoded "username:password"
+
+		value = `{{ printf "{%q:{%q:%q,%q:%q,%q:%q}}" ` +
+			`.Values.kube.registry.hostname ` +
+			`"username" (default "" .Values.kube.registry.username) ` +
+			`"password" (default "" .Values.kube.registry.password) ` +
+			`"auth" (printf "%s:%s" .Values.kube.registry.username .Values.kube.registry.password | b64enc) ` +
+			`| b64enc }}`
```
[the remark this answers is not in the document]
jandubois: Hmm, that is an interesting point. It means we are always assuming "Basic Auth". Maybe we need a mechanism to specify the `auth` credentials in case the registry uses some token auth scheme instead?

As for using `printf` on the literal strings, it just avoids having to escape the doublequotes, I guess. 😄  (They are being printed with `%q` format).

kube/values.go
jandubois: Yeah, but really only if we turn all the functions into settings methods.  Is that what you are suggesting here?

mook-as: Less suggesting and more wishing the world was that way in the first place. I don't think it'd make sense to change it in this PR. 😢

kube/registry_credentials.go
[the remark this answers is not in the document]
jandubois: FWIW, I could not figure out how to use `imagePullSecrets` with token auth. Well, maybe if the token is not expiring, then you could just login with `docker` and then "steal" the token.

Otherwise you would need to somehow deal with refresh tokens.

All references to `imagePullSecrets` I could find seem to be using basic auth.

PR #291, opened 2017-10-26
helm: drop default liveness probe

mook-as:
It wasn't doing anything useful for us; since the pre-start scripts run before monit comes up, that taking too long brings down the pod (and often breaks migrations).

The role manifest still could have explicit liveness probes.

### DIFF
--- a/helm/config.go
+++ b/helm/config.go
@@ -287,8 +287,16 @@ func NewMapping(values ...interface{}) *Mapping {
 	return mapping
 }
 
-// Add a single named node at the end of the list.
+// Add a single named node at the end of the list, replacing any existing
+// entries of the same name.
 func (mapping *Mapping) Add(name string, value interface{}, modifiers ...NodeModifier) {
+	for i, node := range mapping.nodes {
+		if node.name == name {
+			mapping.nodes = append(mapping.nodes[:i], mapping.nodes[i+1:]...)
+			// There can be at most one node with this name
+			break
+		}
+	}
 	mapping.nodes = append(mapping.nodes, namedNode{name: name, node: NewNode(value, modifiers...)})
 }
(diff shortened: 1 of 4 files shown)


helm/config.go
```diff
@@ -287,8 +287,16 @@ func NewMapping(values ...interface{}) *Mapping {
...
+// entries of the same name.
 func (mapping *Mapping) Add(name string, value interface{}, modifiers ...NodeModifier) {
+	for i, node := range mapping.nodes {
+		if node.name == name {
+			mapping.nodes = append(mapping.nodes[:i], mapping.nodes[i+1:]...)
+			// There can be at most one node with this name
+			break
+		}
+	}
 	mapping.nodes = append(mapping.nodes, namedNode{name: name, node: NewNode(value, modifiers...)})
```
jandubois: Seems unrelated to commit message, so should ideally be in a separate commit, but ok because it is small enough and self-explanatory.

kube/pod.go
```diff
@@ -272,31 +269,17 @@ func getContainerLivenessProbe(role *model.Role) (helm.Node, error) {
...
-		probe, complete, err = configureContainerProbe(role, "liveness", role.Run.HealthCheck.Liveness)
+		probe, _, err = configureContainerProbe(role, "liveness", role.Run.HealthCheck.Liveness)
 
 		if probe.Get("initialDelaySeconds").String() == "0" {
 			probe.Add("initialDelaySeconds", defaultInitialDelaySeconds)
 		}
-		if complete || err != nil {
-			return probe, err
-		}
-	}
```
jandubois: I think this check for `complete` should remain, so that we don't emit an empty liveness probe into the YAML.

PR #295, opened 2017-11-07
Add annotations to various kube objects

alexjh:
Adds an `object-annotations` section to the role manifest that will be written to the Kube object's `metadata/annotations`.


alexjh:
I tested this by adding:

```yaml
object-annotations:
  # Used for helm hooks
  key1: value1
  key2: |
    just some random free form text
    we can put whatever we want here
  "helm.sh/hook": post-install
  "helm.sh/hook-weight": "-5"
  "helm.sh/hook-delete-policy": hook-succeeded
```
to the `run` section of `post-deployment-setup`.

kube/utils.go
```diff
@@ -116,9 +116,12 @@ func newTypeMeta(apiVersion, kind string, modifiers ...helm.NodeModifier) *helm.
 	return mapping
 }
 
-func newObjectMeta(name string) *helm.Mapping {
+func newObjectMeta(name string, annotations *map[string]string) *helm.Mapping {
```
mook-as: Annotations don't make sense for most types of objects; I think it would be better to leave it out of `newObjectMeta` and just have it in `newKubeConfig`. Or maybe even leave it out of that, and do it in the callers.

alexjh: Ok, added it just for `jobs` for now, which is the only place we need annotations AFAIK.

Rebased on master and pushed again.

kube/job_test.go
```diff
@@ -130,3 +130,55 @@ func TestJobPostFlight(t *testing.T) {
...
+		metadata:
+			name: role
+			annotations:
+				helm.sh/hook: post-install
+	`, "\t", "    ", -1)
+	if !assert.NoError(yaml.Unmarshal([]byte(expectedYAML), &expected)) {
+		return
+	}
+
+	/*
```
mook-as: What's with this chunk of commented out stuff (that we don't actually care about right now)?

alexjh: Ugh, leftover from the structure I copied. :frowning: 

Fixed in bffb7e7.

PR #300, opened 2017-11-17
Remove secret generation from kube configs

alexjh:
Secret generation is done as a job now. This allows them to be kept across upgrades instead of regenerating them.


alexjh:
@mook-as I hate to type this but: I've added a card to our board for tests, I don't have time before the release to get it done. :crying_cat_face:

mook-as:
Nah, it's okay, I sort of expected that to happen. I'm really just wanting for the SCF/UAA stuff to get sorted out before merging this (otherwise if we needed any _other_ fissile changes it'll break all the things).

PR #301, opened 2017-11-24
Revert "Only skip password types for now"

alexjh:
This reverts commit df3f26631e2b10ed5b51d4289df99b35a402b7ac.

This allows `scf-secret-generator` to update certs and SSH keys as well.


cmd/build-kube.go
```diff
@@ -97,5 +100,12 @@ func init() {
 		"Additional information to use in computing the image tags",
 	)
 
+	buildKubeCmd.PersistentFlags().BoolP(
+		"use-secrets-generator",
+		"",
+		false,
+		"The project uses the a secret generator pre-flight job",
```
alexjh: `s/the a/the` ?

cmd/build-helm.go
```diff
@@ -97,5 +100,12 @@ func init() {
 		"Additional information to use in computing the image tags",
 	)
 
+	buildHelmCmd.PersistentFlags().BoolP(
+		"use-secrets-generator",
+		"",
+		false,
+		"The project uses the a secret generator pre-flight job",
```
alexjh: `s/the a/the`

cmd/build-kube.go
alexjh: `s/the a/the`

kube/stateful_set.go
```diff
@@ -30,6 +30,12 @@ func NewStatefulSet(role *model.Role, settings ExportSettings) (helm.Node, helm.
 	spec := helm.NewMapping()
 	spec.Add("serviceName", fmt.Sprintf("%s-set", role.Name))
 	spec.Add("template", podTemplate)
+	// "updateStrategy" is new in kube 1.7, so we don't add anything to non-helm configs
+	// The default behaviour is "OnDelete"
```
mook-as: This feels like it belongs in a different PR (it's not directly related to secret generation)

jandubois: I agree; it was only merged in here because this integration branch was a substitute for `master` until we had restored backwards compatibility. Do you think this is important enough to split out into a separate PR. It is already in a separate commit: https://github.com/SUSE/fissile/pull/301/commits/4a2a4f866640c4d5d7d5d8fe96be86a4ad5d2b68, which points back to its own Trello card...

cmd/build-helm.go
```diff
@@ -97,5 +100,12 @@ func init() {
 		"Additional information to use in computing the image tags",
 	)
 
+	buildHelmCmd.PersistentFlags().BoolP(
+		"use-secrets-generator",
+		"",
+		false,
+		"The project uses a secret generator pre-flight job",
```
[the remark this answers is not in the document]
jandubois: I have problems creating a short description what it does beyond "preserve the old functionality before we had secrets generator roles".

Some of the things it does:

* no longer generates passwords in the kube secrets
* doesn't include default settings for generated secrets other than passwords (SSH keys and SSL certs and CAs) in `values.yaml`
* calls the secret object `secrets-update` instead of `secrets` but still references the `secret` object from all other objects

I would rather document these requirements in the `README.md` of the `fissile-secrets-generator` repo. I think it is too much detail for a CLI option summary.

But anyways, what would be a better name for this option?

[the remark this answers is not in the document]
jandubois: > doesn't include default settings for generated secrets other than passwords

Ok, I can barely understand what I wrote myself; what I meant was:

* includes the names of generated secrets in `values.yaml`, but ignores the values provided by `--defaults-file`
* this includes the names of generated password, which aren't user-overridable without this option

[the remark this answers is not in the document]
mook-as: Best name I can think of is `--skip-generated-secrets`, but that's not _really_ better.

I think I'm okay with leaving it as `--use-secrets-generator` if the help text explains how to use it (or point to a place that tells you how to use it, given the verbosity).

kube/stateful_set.go
mook-as: 😑 Alright, I guess.

PR #306, opened 2017-12-08
Use routing table to determine container IP address

jandubois:
Using `hostname -i` relies on DNS, which doesn't seem to work reliably on all configurations:

```
routing-api/0:/# hostname -i
hostname: Name or service not known
```


jandubois:
This commit has been cherry-picked into #305 to aid in its testing on Jenkins

PR #307, opened 2017-12-13
Append random suffix to Job names

jandubois:
We want jobs to run again during `helm upgrade`. This means we have to delete the job before creating it again, which helm doesn't support, except for helm "hooks".  Using a random suffix creates a new Job that will still run while the old one can be removed because it is no longer included in the chart.


jandubois:
Replaced by #308

PR #314, opened 2018-01-02
Add a --output-graph option to dump graphviz file for hashes involved

mook-as:
This is to help debugging cases where things were being unexpectedly rebuilt (or not being rebuilt when it should be).

Also adds a bin/graph-diff to examine the differences between two such graphs, colour-coding the differences.

The best way to look at the graphs are probably to generate them via `fdp` (from GraphViz) and view them with `ZGRViewer`.


app/fissile.go
```diff
@@ -1101,3 +1103,66 @@ func (f *Fissile) generateKubeRoles(settings kube.ExportSettings) error {
...
+		return err
+	}
+	if err := f.graphFile.Close(); err != nil {
+		return err
+	}
+	f.graphFile = nil
+	return nil
+}
+
+// GraphNode adds a node to the hash debugging graph
```
andreas-kupries: Should possibly note that this makes the `fissile` struct an implementation of the `ModelGrapher` interface.

app/fissile.go
```diff
@@ -1101,3 +1103,66 @@ func (f *Fissile) generateKubeRoles(settings kube.ExportSettings) error {
...
+	for k, v := range attrs {
+		attrString += fmt.Sprintf("[%s=\"%s\"]", k, v)
+	}
+	if _, err := fmt.Fprintf(f.graphFile, "\"%s\" %s\n", nodeName, attrString); err != nil {
+		return err
+	}
+	return nil
+}
+
+// GraphEdge adds an edge to the hash debugging graph
```
andreas-kupries: See above.

model/roles.go
```diff
@@ -696,51 +707,94 @@ func (r *Role) GetRoleDevVersion(opinions *Opinions, tagExtra, fissileVersion st
 		sort.Strings(keys)
 
 		// ... then add them and their values to the hash precursor
+		// For the graph output, adding all properties individually results in
+		// too many nodes and makes graphviz fall over. So use the hash of them
+		// all instead.
+		propertyHasher := sha1.New()
```
andreas-kupries: Should this be conditional on `grapher != nil` ?

model/roles.go
```diff
@@ -696,51 +707,94 @@ func (r *Role) GetRoleDevVersion(opinions *Opinions, tagExtra, fissileVersion st
...
 		for _, property := range keys {
 			value := flatProps[property]
 			signatures = append(signatures, property, value)
+			propertyHasher.Write([]byte(property))
+			propertyHasher.Write([]byte{0x1F})
+			propertyHasher.Write([]byte(value))
+			propertyHasher.Write([]byte{0x1E})
```
andreas-kupries: Ditto for this block (717-720) ?

model/roles.go
```diff
@@ -696,51 +707,94 @@ func (r *Role) GetRoleDevVersion(opinions *Opinions, tagExtra, fissileVersion st
...
+		}
+		extraGraphEdges = append(extraGraphEdges, []string{
+			fmt.Sprintf("properties/%s:", roleJob.Name),
+			hex.EncodeToString(propertyHasher.Sum(nil))})
```
andreas-kupries: Ditto here ?

app/fissile.go
mook-as: Comment added in e1f5e2f

app/fissile.go
mook-as: Comment added in e1f5e2f

model/roles.go
mook-as: I skipped this one because it's not doing much work, and adding the check complicates the flow a bit (for reading).

model/roles.go
mook-as: Conditional added in e1f5e2f

model/roles.go
mook-as: Conditional added in e1f5e2f

model/roles.go
andreas-kupries: Ok

andreas-kupries:
Merging now. Have graph output, diffing two outputs works, visualization works as well, although meh with ZGRViewer.

PR #317, opened 2018-01-10
Add auth-type option

alexjh:
Add a command line option to set the auth type in the generated
`values.yaml` file. The default remains `~`.

### DIFF
--- a/cmd/build-helm.go
+++ b/cmd/build-helm.go
@@ -14,6 +14,7 @@ var (
 	flagBuildHelmUseMemoryLimits     bool
 	flagBuildHelmTagExtra            string
 	flagBuildHelmUseSecretsGenerator bool
+	flagBuildHelmAuthType            string
 )
 
 // buildHelmCmd represents the helm command
@@ -29,6 +30,7 @@ var buildHelmCmd = &cobra.Command{
 		flagBuildHelmTagExtra = buildHelmViper.GetString("tag-extra")
 		flagBuildHelmUseSecretsGenerator = buildHelmViper.GetBool("use-secrets-generator")
 		flagBuildOutputGraph = buildViper.GetString("output-graph")
+		flagBuildHelmAuthType = buildViper.GetString("auth-type")
 
 		err := fissile.LoadReleases(
 			flagRelease,
@@ -61,6 +63,7 @@ var buildHelmCmd = &cobra.Command{
 			CreateHelmChart:     true,
 			UseSecretsGenerator: flagBuildHelmUseSecretsGenerator,
 			TagExtra:            flagBuildHelmTagExtra,
+			AuthType:            flagBuildHelmAuthType,
 		}
 
 		if flagBuildOutputGraph != "" {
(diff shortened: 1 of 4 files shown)


cmd/build-helm.go
```diff
@@ -29,6 +30,7 @@ var buildHelmCmd = &cobra.Command{
 		flagBuildHelmTagExtra = buildHelmViper.GetString("tag-extra")
 		flagBuildHelmUseSecretsGenerator = buildHelmViper.GetBool("use-secrets-generator")
 		flagBuildOutputGraph = buildViper.GetString("output-graph")
+		flagBuildHelmAuthType = buildViper.GetString("auth-type")
```
mook-as: Shouldn't that be `buildHelmViper` instead of `buildViper` here?

kube/values_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+func TestMakeValues(t *testing.T) {
+	t.Parallel()
+
+	assert := assert.New(t)
+	outDir, err := ioutil.TempDir("", "fissile-generate-auth-")
+	require.NoError(t, err)
+	defer os.RemoveAll(outDir)
+
+	t.Run("Check Default Registry", func(t *testing.T) {
```
mook-as: This one can probably be run in parallel? Actually, most of the other sub-cases too… That is, parallel across the various sub-tests; the toplevel test is already parallel.

kube/values_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	t.Run("Check Custom Auth", func(t *testing.T) {
+		settings := ExportSettings{
+			OutputDir: outDir,
+			RoleManifest: &model.RoleManifest{Roles: model.Roles{},
+				Configuration: &model.Configuration{},
+			},
+		}
+
+		settings.AuthType = "rbac"
```
mook-as: It might be useful to set this to an invalid value to ensure that we take arbitrary values. And maybe use an extra variable / constant for this to make the relationship clear.

kube/values_test.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+
+		node, err := MakeValues(settings)
+
+		assert.NotNil(node)
+		assert.NoError(err)
+
+		auth := node.Get("kube").Get("auth")
+
+		assert.Equal(auth.String(), "~")
```
mook-as: Maybe add a message (extra parameter to `assert.Equal()`) along the lines of "default authorization mode should be nil".

cmd/build-helm.go
alexjh: Fixed in 6d25908

kube/values_test.go
alexjh: Fixed in 7a195d7

kube/values_test.go
alexjh: Fixed in b200a48

kube/values_test.go
alexjh: Fixed in facadbd

PR #318, opened 2018-01-11
`build cleancache`: update to support new path with stemcell hash

mook-as:


app/fissile.go
```diff
@@ -390,7 +390,11 @@ func (f *Fissile) CleanCache(targetPath string) error {
 			continue
 		}
 
-		f.UI.Printf("- Removing %s\n", color.YellowString(key))
+		relpath, err := filepath.Rel(targetPath, cache)
```
andreas-kupries: What is `cache` here ? Should it be `key` instead ?

andreas-kupries: Seems to work however, based on my quick test.
Of course, it semi-smashes an old-style compilation cache a bit (It sees the `hash/compiled` etc directories).
That is not an issue I believe, these things will get regenerated with an old fissile, and not by the newer. Both proper behaviour.

mook-as: Cache is the thing we're iterating (the `/*/*` paths), which is the thing we will be deleting. We're showing that relative to the work directory so it's clearer which copy of the compiled artifact (including which stemcell) we're deleting.

PR #323, opened 2018-02-09
sWAKVtIS Emit references to versioned main secrets.

andreas-kupries:
Ref: https://trello.com/c/sWAKVtIS/551-3-use-versioned-secrets-to-avoid-race-condition
For all related PRs (fissile, scf-secret-generator, uaa-fissile-release, scf)

### DIFF
--- a/kube/pod.go
+++ b/kube/pod.go
@@ -218,7 +218,9 @@ func getEnvVars(role *model.Role, defaults map[string]string, secrets SecretRefM
 		}
 
 		if config.Secret {
-			secretKeyRef := helm.NewMapping("key", secrets[config.Name].Key, "name", secrets[config.Name].Secret)
+			secretKeyRef := helm.NewMapping(
+				"key", secrets[config.Name].Key,
+				"name", secrets[config.Name].Secret+"-{{ .Release.Revision }}")
 
 			envVar := helm.NewMapping("name", config.Name)
 			envVar.Add("valueFrom", helm.NewMapping("secretKeyRef", secretKeyRef))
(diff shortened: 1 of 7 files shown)


kube/pod.go
```diff
@@ -218,7 +218,9 @@ func getEnvVars(role *model.Role, defaults map[string]string, secrets SecretRefM
 		}
 
 		if config.Secret {
-			secretKeyRef := helm.NewMapping("key", secrets[config.Name].Key, "name", secrets[config.Name].Secret)
```
jandubois: You need to wrap the new code into a `	if settings.CreateHelmChart { }` block and keep the old code in the `else` clause because the plain kube spec files don't have helm template expansion.

Also all the test changes need to be reverted.
`

andreas-kupries: Ok. Oops.

andreas-kupries: Handled with commit f670431aa64e514ac17c59006f764cef44442fce

kube/pod.go
```diff
@@ -26,7 +26,7 @@ func NewPodTemplate(role *model.Role, settings ExportSettings, grapher util.Mode
 		return nil, fmt.Errorf("Role %s has no run information", role.Name)
 	}
 
-	vars, err := getEnvVars(role, settings.Defaults, settings.Secrets, settings)
+	vars, err := getEnvVars(role, settings.Defaults, settings)
```
jandubois: It seems weird that we are passing in `settings.Defaults` when we are also passing in `settings`. Given that this is the *only* call to this function (outside of tests), I think we should simplify to just pass in the `settings` object and let the function pick out the defaults itself.

andreas-kupries: Will simplify.

andreas-kupries: Addressed with commit 3fddb6f62a5f36b480ded5064693ef967ff36c86

PR #324, opened 2018-02-14
Run BOSH drain scripts

mook-as:
This adds a Kubernetes pre-stop script which implements the [BOSH job stop lifecycle](https://bosh.io/docs/job-lifecycle.html#stop), including [drain scripts](https://bosh.io/docs/drain.html).  Note that this doesn't extend the time out for shut down, so it may terminate prematurely (but at least it does _something_).

There were some cleanups to `run.sh` too because I thought that was where part of the logic was going to live (it ended up being a dead end).

### DIFF
--- a/scripts/dockerfiles/pre-stop.sh
+++ b/scripts/dockerfiles/pre-stop.sh
@@ -11,49 +11,51 @@
 exec 1> /proc/1/fd/1
 exec 2> /proc/1/fd/2
 
-set -o errexit
+if test -n "${1:-}" ; then
+    # This script is being run from itself, to spawn one drain script
+    # This is necesary to get the parallel execution
+    if ! [ -x "/var/vcap/jobs/$1/bin/drain" ] ; then
+        exit 0
+    fi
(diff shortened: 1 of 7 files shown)


scripts/dockerfiles/pre-stop.sh
```diff
@@ -0,0 +1,60 @@
...
+                sleep $(( 0 - output ))
+            else
+                sleep "${output}"
+                break
+            fi
+        done
+    }
+    {{ range $job := .role.RoleJobs }}
+        drain_job {{ $job.Name }}
+    {{ end }}
```
andreas-kupries: You seem to run the drain scripts sequentially.
The https://bosh.io/docs/job-lifecycle.html#stop claims that BOSH runs them in parallel and then waits for all to finish.
That should be possible for us as well.

mook-as: Fixed in 793747b8f7a431d7287d96c8efcbe93906864b9f

PR #326, opened 2018-02-22
Allow multiple IPs for external IP

mook-as:
This exposes the list nature of the external IP to the configuration, so that in setups where multiple IPs are desired (e.g. for a HA-ish setup) that can be used.

The old-style single-IP configuration is still supported

### DIFF
--- a/kube/service.go
+++ b/kube/service.go
@@ -104,11 +104,20 @@ func NewClusterIPService(role *model.Role, headless bool, public bool, settings 
 		}
 	}
 	if public {
-		externalIP := "192.168.77.77"
+		externalIP := "[ 192.168.77.77 ]"
 		if settings.CreateHelmChart {
-			externalIP = "{{ .Values.kube.external_ip | quote }}"
+			// Backwards compatibility: If .kube.external_ips doesn't exist,
+			// use .kube.external_ip instead (as the single address)
+			externalIP = strings.Replace(`{{
+				default
+					( append .Values.kube.external_ips
+						( .Values.kube.external_ip )
+					)
+					.Values.kube.external_ips
+				| toJson
+			}}`, "\n", " ", -1)
 		}
-		spec.Add("externalIPs", helm.NewList(externalIP))
+		spec.Add("externalIPs", externalIP)
 	}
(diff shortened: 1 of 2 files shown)


mook-as:
Forgot to test upgrades; please don't merge this yet.

mook-as:
Upgrades seem to have no issues for me.

alexjh:
This LGTM but I'm unsure how I should test the backwards compatibility. I can edit the generated `values.yaml` but I don't know if that's what the real scenario is. For the vagrant box, we're generating it, so I don't see how there would be a mismatch. Is the customer reusing their old values.yaml a valid scenario?

What happens if they use our new file but have an old `--set kube.external_ip=1.2.3.4` in the command line of helm? Is that something we need to handle?

I naively changed:
```
vagrant@vagrant-kube:~/scf> diff -u output/helm/values.yaml{.orig,}
--- output/helm/values.yaml.orig 2018-02-23 14:48:34.454369557 -0500
+++ output/helm/values.yaml 2018-02-23 14:45:31.979082455 -0500
@@ -933,8 +933,7 @@
 services:
   loadbalanced: false
 kube:
-  external_ips: 
-  - "192.168.77.77"
+  external_ip: "192.168.77.77"
   storage_class:
     persistent: "persistent"
     shared: "shared"
```

But `make run` returned an error:
```
Error: render error in "cf-opensuse/templates/tcp-router.yaml": template: cf-opensuse/templates/tcp-router.yaml:237:53: executing "cf-opensuse/templates/tcp-router.yaml" at <.Values.kube.externa...>: wrong type for value; expected []interface {}; got interface {}
```

mook-as:
I believe the correct edit is:
- In `values.yaml`, have `external_ips: []`
- In the custom config file, have `external_ip: 192.168.77.77`

Alternatively, say `--set kube.externap_ip=192.168.77.77` on the command line.  If you want to do the new thing, it's `--set kube.external_ips[0]=192.168.77.77`.

But that _does_ mean we should remove the default value in the `external_ips`, though.

alexjh:
Ok, LGTM. I assume there will be a corresponding change to the `make/run` scripts in UAA and SCF to set the external IP when fissile is bumped there.

PR #328, opened 2018-02-28
run.sh: abort waiting for monit if monit is missing

mook-as:
If monit has already exited, it can't possibly respond to queries about if the processes are still running.  In that case, just stop asking.


alexjh:
:+1: I don't see a card for this, I'll test it now.

alexjh:
LGTM, everything exits quickly again, other than diego which took the longest.

PR #329, opened 2018-03-02
Extend fissile to generate proper memory requests and limits.

andreas-kupries:
Ref: https://trello.com/c/UEvqSutH/576-sane-memory-limits-for-all-pods

See also SCF branch https://github.com/SUSE/scf/tree/aku/UEvqSutH-mem-limits
where the role-manifest is modified based on memory measurements for the SCF roles.

### DIFF
--- a/kube/values.go
+++ b/kube/values.go
@@ -38,11 +38,14 @@ func MakeValues(settings ExportSettings) (helm.Node, error) {
 	}
 	env.Sort()
 
-	sizing := helm.NewMapping(
-		"HA", false,
-		"memory", helm.NewMapping("slack", settings.MemLimitFactor,
-			"requests", false,
-			"limits", false))
+	memSizing := helm.NewMapping()
+	memSizing.Add("limit_factor", settings.MemLimitFactor, helm.Comment("Factor to calculate the limits from the requests\nMust be 1 or larger."))
+	memSizing.Add("requests", false, helm.Comment("Flag to activate memory requests"))
+	memSizing.Add("limits", false, helm.Comment("Flag to activate memory limits"))
+
+	sizing := helm.NewMapping()
+	sizing.Add("HA", false, helm.Comment("Flag to activate high-availability mode"))
+	sizing.Add("memory", memSizing, helm.Comment("Global memory configuration"))
 
 	for _, role := range settings.RoleManifest.Roles {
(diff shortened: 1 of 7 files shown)


kube/values.go
```diff
@@ -38,7 +38,12 @@ func MakeValues(settings ExportSettings) (helm.Node, error) {
 	}
 	env.Sort()
 
-	sizing := helm.NewMapping("HA", false)
+	sizing := helm.NewMapping(
+		"HA", false,
+		"memory", helm.NewMapping("slack", settings.MemLimitFactor,
```
jandubois: I don't like the term `slack` here; in my mind slack is something you add, not multiply. 😄

I would also prefer that we don't use different terms internally and externally, so it should be called `limit-factor` by default (or if we use a different user term, then the internal name should be updated to match).

And whatever we choose, we should also add a comment to it, explaining exactly what it does.

(and if you are feeling generous, add a comment to the `HA` setting as well).

jandubois: And by comment I mean a comment in the `values.yaml` file, not in the fissile source code. 😄

andreas-kupries: sure.

andreas-kupries: Addressed with commit d7d1dd7ccb049ac3889562f1a7daa511bc7b3160
(Note, using `limit_factor` because the dash chokes kube.)

andreas-kupries:
The sprig function `mul` takes to two ints and returns int.
(Per https://godoc.org/github.com/Masterminds/sprig)

jandubois:
> The sprig function mul takes to two ints and returns int.

Ah, ok, never mind then. I thought sprig had support for floating point math.

andreas-kupries:
Well, I went by the docs. We could try and see what happens for float values.

andreas-kupries:
Ok, calling it factor and will remove the redundant phrasing.

andreas-kupries:
And done with commit 9e86604c783f3bf3ef96134e2ad992d3f5694d8d

jandubois:
> Well, I went by the docs. We could try and see what happens for float values.

Don't bother; the docs are correct: https://github.com/Masterminds/sprig/blob/ded3565ee06df6d916eb8b0a5815c38c3494a20c/functions.go#L179-L185

andreas-kupries:
Ok. Thanks for looking.

PR #330, opened 2018-03-08
Templatize tcp routing ports

jandubois:


kube/service.go
```diff
@@ -108,14 +126,14 @@ func NewClusterIPService(role *model.Role, headless bool, public bool, settings
 		if settings.CreateHelmChart {
 			// Backwards compatibility: If .kube.external_ips doesn't exist,
 			// use .kube.external_ip instead (as the single address)
-			externalIP = strings.Replace(`{{
+			externalIP = strings.Join(strings.Fields(`{{
```
andreas-kupries: This seems to have nothing to to with templatized ports.

jandubois: > This seems to have nothing to to with templatized ports.

It doesn't, which is why it is in a commit by itself, and the commit doesn't have the Trello card name prefix.

It is just a tiny code cleanup thing that seems small enough to not need a separate PR.

But we can split it out if you want. In which case we should probably move your cleanup of `helm/config.go` into a separate PR as well.

andreas-kupries: Ok. Accepted.

andreas-kupries:
Some further notes
- Tweaked values.yml to use range 30000-30019 for external port of tcp routing.
- Cluster start :ballot_box_with_check: 
- Brain test :boom: (2 test, tcp-routing, and cf-usb, which uses it as well) `30007: Connection refused`

I don't see this as an issue with `fissile` and the change.
It likely is something in SCF, where it still depends on the original 20000-20009 range, somehow.
When I ran a test which simply extended the range to 20019, keeping the start at 20000
it was :ballot_box_with_check: The issue happens when the range is moved off 20000.

... Talked with Jan ... Apparently we need properties which define the proper range of internal ports.

So, requesting changes.



I better make a ticket for that.

kube/service.go
```diff
@@ -55,31 +55,49 @@ func NewClusterIPServiceList(role *model.Role, headless, private bool, settings
...
+			block := fmt.Sprintf("range $port := until (int %s.count)", sizing)
+			newPort := helm.NewMapping()
+			newPort.Set(helm.Block(block))
+
+			portName := port.Name
+			if port.Max > 1 {
+				portName = fmt.Sprintf("%s-{{ $port }}", portName)
+			}
+			newPort.Add("name", portName)
+			newPort.Add("port", fmt.Sprintf("{{ add (int %s.port) $port }}", sizing))
```
andreas-kupries: Another change required.

Without it we geta very obscure render error: `can't evaluate field Values in type int`. 
Tracked it to https://github.com/kubernetes/helm/issues/1311 (
Inside the `range` scoping has changed. I.e. `.` (dot) is not the toplevel, but the same as `$port` (See Go templating, https://golang.org/pkg/text/template/#pkg-overview, range Action). As an `int` it does not have a `Values` field.

Fix is to use `$` as the unchanging top of the environment, i.e `"{{ add (int $%s.port) $port }}"`
The `.` needed after `$` is, I believe, part of `sizing`.

The output in `tcp-router.yaml` should look like 
```
port: {{ add (int $.Values.sizing.tcp_router.ports.tcp_route.port) $port }}
```

There are two places in the output (the two services of tcp-router) where I see the issue, fix is the same for both.
Not sure if both are handled by this code here. If yes we are good.

That said, even with this change the relevant brains tests (004, 008) still go down in flames.
```Failed to connect to tcp.cf-dev.io port 20017: Connection refused```

The config might be ok:
Service `tcp-router` has:
```
Port:              tcp-route-0  30000/TCP
TargetPort:        %!d(string=tcp-route-0)/TCP
Endpoints:         172.16.0.232:20000
```

Service `tcp-router-public` shows:
```
Port:              tcp-route-0  30000/TCP
TargetPort:        %!d(string=tcp-route-0)/TCP
Endpoints:         172.16.0.232:20000
```

Pod `tcp-router`:
```
    Ports:          80/TCP, 20000/TCP, 20001/TCP, 20002/TCP, 20003/TCP, 20004/TCP, 20005/TCP, 20006/TCP, 20007/TCP, 20008/TCP, 20009/TCP, 20010/TCP, 20011/TCP, 20012/TCP, 20013/TCP, 20014/TCP, 20015/TCP, 20016/TCP, 20017/TCP, 20018/TCP
```

What I am wondering about is that the tcp-routing service seems to hand out the internal port numbers.

And now I am wondering further if for tcp-routing `count` is the only thing which should be user-customizable, that internal and external port must be the same for this set, because of the above, the internal ports being handed out by the system.

PR #331, opened 2018-03-08
Implement CPU limits in fissile. Analogous to memory limits.

andreas-kupries:
Ref: https://trello.com/c/6925zx7v/584-2-cpu-being-able-to-enable-cpu-limits-whilst-leaving-them-off-by-default

Both cpu and memory request and limits got redone.
For the spec worked from see the ref above, top comments from Mar 9.

* The limit-factors are gone. Role manifest specifies both requests and limits, no implicit coupling.
* By default all requests and limits are globally off in the `values.yaml`.
* The inactive defaults are inherited from the manifest.
* It is possible to set a request or limit for a specific role to `~` , i.e. null and the role will disable this when starting the role. The template checks both global flag and per-role information before activating a request or limit (`if and ...`).

Currently tested against SCF, with the memory limits.
Inactive, globally active, and globally active with mysql request disabled.

### DIFF
--- a/cmd/build-kube.go
+++ b/cmd/build-kube.go
@@ -12,7 +12,7 @@ var (
 	flagBuildKubeOutputDir       string
 	flagBuildKubeDefaultEnvFiles []string
 	flagBuildKubeUseMemoryLimits bool
-	flagBuildKubeMemLimitFactor  int
+	flagBuildKubeUseCPULimits    bool
 	flagBuildKubeTagExtra        string
 )
 
@@ -26,7 +26,7 @@ var buildKubeCmd = &cobra.Command{
 		flagBuildKubeOutputDir = buildKubeViper.GetString("output-dir")
 		flagBuildKubeDefaultEnvFiles = splitNonEmpty(buildKubeViper.GetString("defaults-file"), ",")
 		flagBuildKubeUseMemoryLimits = buildKubeViper.GetBool("use-memory-limits")
-		flagBuildKubeMemLimitFactor = buildKubeViper.GetInt("mem-limit-factor")
+		flagBuildKubeUseCPULimits = buildKubeViper.GetBool("use-cpu-limits")
 		flagBuildKubeTagExtra = buildKubeViper.GetString("tag-extra")
 		flagBuildOutputGraph = buildViper.GetString("output-graph")
 
@@ -56,7 +56,7 @@ var buildKubeCmd = &cobra.Command{
 			Organization:        flagDockerOrganization,
 			Repository:          flagRepository,
 			UseMemoryLimits:     flagBuildKubeUseMemoryLimits,
-			MemLimitFactor:      flagBuildKubeMemLimitFactor,
+			UseCPULimits:        flagBuildKubeUseCPULimits,
 			FissileVersion:      fissile.Version,
 			Opinions:            opinions,
 			CreateHelmChart:     false,
@@ -105,11 +105,11 @@ func init() {
 		"Include memory limits when generating kube configurations",
 	)
 
-	buildKubeCmd.PersistentFlags().IntP(
-		"mem-limit-factor",
+	buildKubeCmd.PersistentFlags().BoolP(
+		"use-cpu-limits",
 		"",
-		3,
-		"Factor for determining limits from requests",
+		true,
+		"Include cpu limits when generating helm chart",
 	)
 
 	buildKubeCmd.PersistentFlags().StringP(
--- a/kube/deployment.go
+++ b/kube/deployment.go
@@ -109,10 +109,6 @@ func replicaCheck(role *model.Role, controller *helm.Mapping, service helm.Node,
 		controller.Add("_oddReplicas", fail, helm.Block(block))
 	}
 
-	failLimitFactor := fmt.Sprintf(`{{ fail "The memory limit factor must be at least 1" }}`)
-	blockLimitFactor := fmt.Sprintf("if lt (int .Values.sizing.memory.limit_factor) 1")
-	controller.Add("_minLimitFactor", failLimitFactor, helm.Block(blockLimitFactor))
-
 	controller.Sort()
 
 	return nil
--- a/kube/pod.go
+++ b/kube/pod.go
@@ -81,12 +81,12 @@ func NewPodTemplate(role *model.Role, settings ExportSettings, grapher util.Mode
 		} else {
 			if role.Run.CPU != nil {
 				if role.Run.CPU.Request != nil {
-					requests.Add("cpu", fmt.Sprintf("%dm", int(0.5+1000**role.Run.CPU.Request)))
+					requests.Add("cpu", fmt.Sprintf("%dm", int(*role.Run.CPU.Request*1000+0.5)))
 				} else {
 					requests.Add("cpu", nil)
 				}
 				if role.Run.CPU.Limit != nil {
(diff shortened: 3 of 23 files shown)


cmd/build-kube.go
```diff
@@ -112,6 +118,20 @@ func init() {
...
+		"",
+		true,
+		"Include cpu limits when generating helm chart",
+	)
+
+	buildKubeCmd.PersistentFlags().IntP(
+		"cpu-limit-factor",
+		"",
+		2,
+		"Factor for determining limits from requests",
```
viovanov: Can you please improve this description? I don't understand it as-is.

kube/deployment.go
```diff
@@ -109,9 +109,16 @@ func replicaCheck(role *model.Role, controller *helm.Mapping, service helm.Node,
...
-	failLimitFactor := fmt.Sprintf(`{{ fail "The memory limit factor must be at least 1" }}`)
-	blockLimitFactor := fmt.Sprintf("if lt (int .Values.sizing.memory.limit_factor) 1")
-	controller.Add("_minLimitFactor", failLimitFactor, helm.Block(blockLimitFactor))
+	if settings.UseMemoryLimits {
+		failMemLimitFactor := fmt.Sprintf(`{{ fail "The memory limit factor must be at least 1" }}`)
+		blockMemLimitFactor := fmt.Sprintf("if lt (int .Values.sizing.memory.limit_factor) 1")
+		controller.Add("_minMemLimitFactor", failMemLimitFactor, helm.Block(blockMemLimitFactor))
+	}
+	if settings.UseCPULimits {
+		failCPULimitFactor := fmt.Sprintf(`{{ fail "The memory limit factor must be at least 1" }}`)
```
viovanov: The message here should be about CPU, not memory - correct?

kube/pod.go
```diff
@@ -35,19 +35,41 @@ func NewPodTemplate(role *model.Role, settings ExportSettings, grapher util.Mode
...
+		}
+	}
+	if settings.UseCPULimits {
+		if settings.CreateHelmChart {
+			requests.Add("cpu",
+				helm.NewNode(fmt.Sprintf("{{ int .Values.sizing.%s.vcpu_count }}m", roleVarName),
+					helm.Block("if .Values.sizing.cpu.requests")))
 
-			resources = helm.NewMapping("requests", reqMap, "limits", limMap)
+			limits.Add("cpu",
```
viovanov: Can you add a comment explaining what this yields given some input?

viovanov:
I'd like to see some explanation whether this is just about limits, or also guarantees.

i.e.

With limits only, if a user has 8 CPUs, and we have 20 roles, each with a 2000m cpu limit, they can run SCF fine. The CPUs will be shared, but any one of them won't be able to use more than 2 of the 8.

With guarantees, the user will need 40000m CPUs available, which is very different.

viovanov:
Reading more - looks like we're using information for "requests" and making assumptions about limits derived from that. I don't think this is the correct approach for what we're trying to achieve.

cmd/build-kube.go
andreas-kupries: This is currently an integer number used to calculate a limit from the request, i.e
`limit = factor * request` ... Analogous to memory req/limit, except feeling somewhat wrong and not having anything else.

kube/deployment.go
andreas-kupries: Definitely. Thanks for the catch. Will fix.

kube/pod.go
andreas-kupries: You mean how the generated template looks like ? Can do. Has to wait a bit, currently set to testing Jan's port templating, with a different fissile.

kube/pod.go
```diff
@@ -35,19 +35,65 @@ func NewPodTemplate(role *model.Role, settings ExportSettings, grapher util.Mode
...
+					helm.Block(fmt.Sprintf("if and .Values.sizing.memory.limits .Values.sizing.%s.memory.limit", roleVarName))))
 		} else {
-			resources = helm.NewMapping(
-				"requests", helm.NewMapping("memory", fmt.Sprintf("%dMi", role.Run.Memory)),
-				"limits", helm.NewMapping("memory", fmt.Sprintf("%dMi", settings.MemLimitFactor*role.Run.Memory)))
+			if role.Run.Memory != nil {
+				if role.Run.Memory.Request != nil {
+					requests.Add("memory", fmt.Sprintf("%dMi", *role.Run.Memory.Request))
+				} else {
+					requests.Add("memory", nil)
```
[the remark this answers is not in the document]
andreas-kupries: That is the kube output
Leaving the line in serves as an implicit reminder that you can edit this value if needed.

kube/pod.go
```diff
@@ -35,19 +35,65 @@ func NewPodTemplate(role *model.Role, settings ExportSettings, grapher util.Mode
...
+			requests.Add("cpu",
+				helm.NewNode(fmt.Sprintf("{{ int .Values.sizing.%s.cpu.request }}m", roleVarName),
+					helm.Block(fmt.Sprintf("if and .Values.sizing.cpu.requests .Values.sizing.%s.cpu.request", roleVarName))))
+			limits.Add("cpu",
+				helm.NewNode(fmt.Sprintf("{{ int .Values.sizing.%s.cpu.limit }}m", roleVarName),
+					helm.Block(fmt.Sprintf("if and .Values.sizing.cpu.limits .Values.sizing.%s.cpu.limit", roleVarName))))
+		} else {
+			if role.Run.CPU != nil {
+				if role.Run.CPU.Request != nil {
+					requests.Add("cpu", fmt.Sprintf("%dm", int(0.5+1000**role.Run.CPU.Request)))
```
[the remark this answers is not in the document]
andreas-kupries: I will check. IIRC `make format` blared at me when I had whitespace. Maybe reorder the terms.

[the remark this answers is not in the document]
andreas-kupries: Done in commit 51f58a556dd7e9d84accc0ec894c641193a838ce
Had to reorder the terms. Adding whitespace makes `goimports` unhappy.

kube/pod.go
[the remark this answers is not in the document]
alexjh: :+1:

alexjh:
Is it valid to specify a limit without a request? When I tried changing the role manifest:

```
-    memory: 69
+#   memory: 69
     virtual-cpus: 2
+    mem:
+      limit: 300
```

The resulting values.yaml file generated:
```
    # Unit [MiB]
    memory:
      request: 0
      limit: 300
```

From what I can tell, it's using docker to control these limits, and 0 isn't valid according to the docs 
https://docs.docker.com/engine/reference/run/#runtime-constraints-on-resources, although I admit, I haven't tested it.

andreas-kupries:
First, limit without request should be possible. I remember trying that early, when working on the memory stuff, before cpu limits and the big rewrite.

The manifest to values conversion looks to be a bug. Instead of `request: 0` it should be `request: ~`, i.e. nil.

Having a look now.

andreas-kupries:
I see now. As `mem.request` is not specified it falls back to `memory`, and the yaml unmarshaling uses a default of `0` when that is not specified. Making it a pointer should make the fallback `nil` instead.

andreas-kupries:
Reworked fallback handling, see commit dfaec97fdd96e1ca0ea1125509582e9b4e9d2f40

PR #333, opened 2018-03-12
Only use numeric part of minor kube version

jandubois:
Ignore the "+" in "1.8+" on GKE.

### DIFF
--- a/kube/utils.go
+++ b/kube/utils.go
@@ -143,5 +143,6 @@ func minKubeVersion(major, minor int) string {
 	ver := ".Capabilities.KubeVersion"
 	// "Major > major || (Major == major && Minor >= minor)"
 	// The int conversions are necessary because Major/Minor in KubeVersion are strings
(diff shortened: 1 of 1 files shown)


jandubois:
> Do you trust them to not do that for major too?

I don't expect them to do it for major. Granted, I didn't expect it for minor either. But major changes much less frequently. We can deal with it if it ever happens.

> Is there any way we can test this? As in, just evaluate a template (not via helm, but possibly via sprig) and make sure things do what we expect?

@viovanov is going to test it tomorrow on GKE, so maybe just let him review this?

mook-as:
Sorry, by "test" I mean "a test file that gets run on `go test`". But yes, we should test it on a real cluster too :)

jandubois:
> Sorry, by "test" I mean "a test file that gets run on go test

Oh, I see. This would really be a test to see that sprig does what it claims it does. Doesn't really feel right to put here. Where would we stop?

There is also the problem that we currently don't test any of the helm-specific template code *at all*, which is a much bigger issue.

jandubois:
FWIW, this change works as expected in the Vagrant box...

viovanov:
Works on GKE 👍

PR #338, opened 2018-03-20
Redo how secrets are generated

jandubois:
* All user provided secrets are stored in a separate secret called `secrets`.

* Only generated secrets are managed by the secrets generator.

* The user can update `kube.secrets_generation_counter` in `values.yaml` to force secrets rotation. This value is available to the roles via the `KUBE_SECRETS_GENERATION_COUNTER` variable.

* The name of the generated secrets includes both the helm chart version and the generation counter. This name is available to the roles via the `KUBE_SECRETS_GENERATION_NAME` variable.

  All pods will be rotated when the chart is updated (new generated secrets may have been added), or when the user incremented the generation counter.

* All pods have an annotation `checksum/config` whose value is a checksum of all user provided secrets.

  All pods will be restarted whenever the manually specified secrets change.

* When the user specifies a secret that also has a generator, then the chart will point the environment variable to the "secrets" secrets, and the generated value will be ignored.

* All secrets are added to `values.yaml` in a separate `secrets` section and are no longer included under `env`.

* The `secrets-update-#` secret mechanism is no longer necessary and has been removed.

[trello#0s4zlOZg]


kube/pod.go
```diff
@@ -319,20 +325,42 @@ func getEnvVars(role *model.Role, settings ExportSettings) (helm.Node, error) {
...
+			envVar := helm.NewMapping("name", config.Name, "valueFrom", helm.NewMapping("secretKeyRef", secretKeyRef))
+
+			if !settings.CreateHelmChart {
+				env = append(env, envVar)
+			} else {
+				envVar.Set(helm.Block(fmt.Sprintf("if .Values.secrets.%s", config.Name)))
+				env = append(env, envVar)
+
+				secretKeyRef := helm.NewMapping("key", util.ConvertNameToKey(config.Name))
+				secretKeyRef.Add("name", "secrets-{{ .Chart.Version }}-{{ .Values.kube.secrets_generation_counter }}")
```
mook-as: Hmm, we can't jsut add the `Block` on the `name` value, can we. Boo.

kube/pod.go
```diff
@@ -139,7 +139,11 @@ func NewPodTemplate(role *model.Role, settings ExportSettings, grapher util.Mode
 	spec.Sort()
 
 	podTemplate := helm.NewMapping()
-	podTemplate.Add("metadata", newObjectMeta(role.Name))
+	meta := newObjectMeta(role.Name)
+	if settings.CreateHelmChart {
+		meta.Add("annotations", helm.NewMapping("checksum/config", `{{ include (print $.Template.BasePath "/secrets.yaml") . | sha256sum }}`))
```
mook-as: Shouldn't this include the generation number / generated secrets, so that pods get re-rolled when you just bump the generation?

kube/pod.go
jandubois: Yeah, we can't because the name inside the mapping must be unique; the second "name" would replace the first one. :(

jandubois: I did simplify the code a bit in https://github.com/SUSE/fissile/commit/f0b082c0575eef715ec9775a3795ada7c3f4dcef though.

kube/pod.go
```diff
@@ -248,6 +252,21 @@ func getVolumeMounts(role *model.Role) helm.Node {
 	return helm.NewNode(mounts)
 }
 
+var generatedSecretsName = "secrets-{{ .Chart.Version }}-{{ .Values.kube.secrets_generation_counter }}"
```
mook-as: This could be a `const`

kube/pod.go
```diff
@@ -248,6 +252,21 @@ func getVolumeMounts(role *model.Role) helm.Node {
...
+
+func makeSecretVar(name string, generated bool, modifiers ...helm.NodeModifier) helm.Node {
+	secretKeyRef := helm.NewMapping("key", util.ConvertNameToKey(name))
+	if generated {
+		secretKeyRef.Add("name", generatedSecretsName)
+	} else {
+		secretKeyRef.Add("name", "secrets")
```
mook-as: `secrets` might need to be a constant too

kube/pod.go
```diff
@@ -319,20 +340,38 @@ func getEnvVars(role *model.Role, settings ExportSettings) (helm.Node, error) {
...
-			envVar.Add("valueFrom", helm.NewMapping("secretKeyRef", secretKeyRef))
+		if config.Secret {
+			if !settings.CreateHelmChart {
+				env = append(env, makeSecretVar(config.Name, false))
+			} else {
+				if config.Immutable && config.Generator != nil {
+					env = append(env, makeSecretVar(config.Name, true))
+				} else {
+					block := helm.Block(fmt.Sprintf("if not .Values.secrets.%s", config.Name))
+					env = append(env, makeSecretVar(config.Name, true, block))
```
mook-as: This `else` block runs if there's isn't a generator… which wouldn't apply here, right?

kube/secret.go
```diff
@@ -14,42 +14,39 @@ import (
...
-				errString := fmt.Sprintf("%s configuration missing", cv.Name)
-				value = fmt.Sprintf(`{{ required "%s" .Values.env.%s | b64enc | quote }}`, errString, cv.Name)
+			if cv.Generator == nil {
+				errString := fmt.Sprintf("secrets.%s has not been set", cv.Name)
+				value = fmt.Sprintf(`{{ required "%s" .Values.secrets.%s | b64enc | quote }}`, errString, cv.Name)
+				if cv.Immutable {
+					comment += "\nThis value is immutable and must not be changed once set."
+				}
+				data.Add(key, helm.NewNode(value, helm.Comment(comment)))
+			} else if !cv.Immutable {
```
mook-as: What happens if there is a generator and it is immutable?

kube/values.go
```diff
@@ -11,20 +11,19 @@ import (
...
+	generated := helm.NewMapping()
+
 	for name, cv := range model.MakeMapOfVariables(settings.RoleManifest) {
 		if strings.HasPrefix(name, "KUBE_SIZING_") || cv.Type == model.CVTypeEnv {
 			continue
 		}
-		// When not using the secrets-generator, passwords are generated by helm
-		// templates in secret.yaml and are not overridable.
-		if !settings.UseSecretsGenerator && cv.Generator != nil && cv.Generator.Type == model.GeneratorTypePassword {
+		if cv.Immutable && cv.Generator != nil {
```
mook-as: Comment on why this case isn't handled?

kube/pod.go
jandubois: > This could be a const

Yeah, was supposed to be...

kube/pod.go
jandubois: Thanks for catching this, user specified secrets should be loaded unconditionallyf rom "secrets" only.

kube/secret.go
jandubois: It was a conscious decision (by me & @viovanov) to make immutable secrets with a generator not overridable by the user.  There should never be a need to do so. Therefore these secrets don't need to be included in `values.yaml` or `templates/secrets.yaml`.

Personally I think *none* of the generated secrets need to be user-overridable if we have a working rotation mechanism, but @viovanov did not agree, so the rest are still there...

mook-as: Then please add a comment (or better, an assertion) to show that we thought about the case (and rejected it).

Issue #340, opened 2018-03-22
Add support for multi-containers aka sidecar per Pod


qu1queee:
Hi,

We currently have a strong use case for supporting multiple containers per Pod(aka sidecars). By default in Kube, a pod configuration file allows to have multiple entries under `spec.containers`.  

However, the current setup of `fissile`(e.g. running a `fissile build helm`), does not support this by default. We can always patch the YAML templates generated by `fissile`, but ideally we want to support this _sidecar_ feature as part of the `fissile` natural behavior(e.g. throught `fissile build helm`).

We would like to open a discussion on how to achieve this, and then PRing this to you. The main concern is that we will need to modify/introduce some of the YAML syntax for the _role-manifest_ .

Our current idea would be to introduce two new keys in the `role-manifest.yml`, as follows:
```yaml
roles:
- name: nginx
  jobs:
  - name: nginx
    release_name: nginx
  tags:
  - clustered
  sidecars:                 # new list to reference sidecar roles
  - log-forwarder
  run:
    memory: 256
    virtual-cpus: 4
    exposed-ports: [...]
    shared-volumes:
      - path: /mnt/share
        tag: common-share
        empty-dir: true    # new reference for a "volumes" entry in the template YAML
- name: log-forwarder
  type: sidecar            # new type called sidecar
  jobs:
  - name: elk
    release_name: elk
  tags:
  - clustered
  run:
    memory: 256
    virtual-cpus: 4
    exposed-ports: [...]
    shared-volumes:
      - path: /mnt/share
        tag: common-share
        empty-dir: true    # new reference for a "volumes" entry in the template YAML
```

As illustrated above, the   _`roles.nginx.sidecars`_ will ideally be the trigger to colocate in the `nginx` pod template, a second item under `spec.containers` with the properties of the `log-forwarder` sidecar.

Inside the `log-forwarder` role, the  _`roles.component.type`_ will indicate that this Pod YAML template should not be generated, whilst we expect the docker image to be generated.

With this _sidecar_, we would also require a simple volume sharing solution. For example, as defined in the [Creating a Pod that runs two Containers](https://kubernetes.io/docs/tasks/access-application-cluster/communicate-containers-same-pod-shared-volume/) scenario, we just want to share one directory between the containers. For that, an additional `role-manifest.yml` modification would be required to enable support for [`emptyDir: {}`](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir).

What do you think?

Thanks in advance for your feedback.
[...]

PR #341, opened 2018-03-23
Support hostpath volumes

mook-as:
This adds support for hostpath volumes, for SUSE/scf#1345 to mount the external `/sys/fs/cgroup` in `diego-cell`.  It also accepts (but does not implement) emptyDir volumes, sort of on the way to #340 (if we decide to implement this way, rather than `empty-dir: true` as suggested).

This is more for discussion than merging; for one thing, I'm not sure hostpath would always be available.

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -1026,6 +1026,20 @@ func (f *Fissile) generateBoshTaskRole(outputFile *os.File, role *model.Role, se
 	return nil
 }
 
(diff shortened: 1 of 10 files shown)


mook-as:
Rebased, and added the helm value thing (defaulting true). Also tests, because I am not confident enough in myself to not have it.  I was unable to find anything to make that detection automatic.

PR #342, opened 2018-03-26
Restore support for empty lists

jandubois:


jandubois:
> I'd like a test to make sure we don't create empty volume claim template lists

Oh, I agree, but we don't have *any* tests for generated helm charts so far, so there is no good place to put those tests. This is really blocked by https://trello.com/c/42gyM3NL/120-create-fissile-tests-for-helm-charts

mook-as:
Oh poop.

PR #347, opened 2018-04-13
Drop auto-creation of stateful sets for roles with volume decls.

andreas-kupries:
This forces the explicit use of tags at roles to trigger their creation.
Further tweaked make, announce cleaning like all the other ops.

https://trello.com/c/0zylWw07/606-2-fissile-statefulsets-should-be-explicitly-defined

Dependent: https://github.com/SUSE/scf/pull/1456
Dependent: https://github.com/SUSE/uaa-fissile-release/pull/94

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -35,7 +35,7 @@ roles:
   - name: nats
     release_name: nats             # The name of the BOSH release this is from
   tags:
-  - clustered                      # Mark this role as self-clustering
+  - clustered                      # Mark this role as self-clustering => StatefulSet
   run:                             # Runtime configuration
     scaling:                       # Auto-scaling limits
       min: 1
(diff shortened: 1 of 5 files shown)


docs/configuration.md
```diff
@@ -132,7 +132,60 @@ headers (for example, to set the `Accept:` header to request JSON responses).
...
+it as a [StatefulSet].
+
+The second way of causing fissile to do that is to tag a role as
+`indexed`. Any role which requires local storage should be tagged as
+such to take advantage of volume claim templates. An example would be
+the BLOBSTORE of CF. See the example below.
+
+Another difference between `clustered` and `indexed` roles is that
+fissile creates a public service for the latter, providing a single
+point of access to the pods for the role.
```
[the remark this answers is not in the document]
andreas-kupries: @jimmykarily @mook-as 
I have no idea what the answer to the question is.
Which is why not quite a week ago the ticket got a ping to @viovanov as the authority.
So far this ping has not been answered.
I made the change above to at least document the existing behaviour based on what I saw from the a/p redesign work (the table listing resource, service, strategy by role type).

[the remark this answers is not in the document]
jimmykarily: :+1:

andreas-kupries:
Typo fixed with commit 27b9a2e9a7da03d75c796028c3c2cc89c53c04ee

andreas-kupries:
Oh. Needs a rebase.

andreas-kupries:
Rebase done

docs/configuration.md
```diff
@@ -132,7 +132,65 @@ headers (for example, to set the `Accept:` header to request JSON responses).
...
+  - name: blobstore_url_signer
+  tags:
+  - indexed                      # Mark this role as indexed => StatefulSet
+  run:
+    scaling:
+      min: 1
+      max: 1
+      # Not HA capable; use external HA storage instead
+    capabilities: []
+    persistent-volumes:          # Local persistent storage, shared across the set
```
mook-as: This is *not* shared across the set; that's `shared-volumes`.
Also, since #341 landed, it should be written as:
```yaml
volumes:
- path: /var/vcap/store
  tag: blobstore-data
  type: persistent # or type: shared, type: host
  size: 50
```

docs/kubernetes.md
```diff
@@ -35,12 +35,26 @@ cluster that is needed for other purposes.
...
-example, a NATS role would fall under this category.  Secondly, any roles which
-require local storage will be a StatefulSet to take advantage of volume claim
-templates.
+
+Fissile emits a [StatefulSet] under two circumstances.
+
+Any self-clustering roles (i.e. any role with the `clustered` tag)
+will be a StatefulSet, in order for each pod to be addressable (so
+that they can talk to each other). For example, a NATS role would fall
+under this category.
```
mook-as: Except that [NATS is indexed](https://github.com/SUSE/scf/blob/b5e5fd51eccb736bbf8dd9f7d92e802eefc4df7e/container-host-files/etc/scf/config/role-manifest.yml#L101). (Probably for the service in front of it doing LB work)

docs/configuration.md
andreas-kupries: Will fix the `shared` thing.
WRT the 341, did we have SCF and UAA PRs fixing the syntax in the respective role manifests ?

docs/kubernetes.md
andreas-kupries: Ok, so NATs changed at some point and nobody bothered to update the fissile docs (most likely not remembered that it was referenced in them). So, ok, I can do that, i.e. look for a new example of a clustered role. (NATS has the big example either in fise or configuration.md).

docs/configuration.md
mook-as: No, because the old syntax is still supported.

docs/kubernetes.md
andreas-kupries: Done with commit 345848333b28aa12daf75028b9ad103163fe782c.
Kept NATS, rewrote to `indexed`, then went with `mysql` for a clustered role.

docs/configuration.md
andreas-kupries: Ok. So I don't have to fix up my SCF/UAA work in that regard either.
Reworked the example. Commit 345848333b28aa12daf75028b9ad103163fe782c.

PR #351, opened 2018-05-04
Enable operator overrides for role capabilities

andreas-kupries:
Ref https://trello.com/c/BxUbBojW/642-5-eks-label-for-specific-pods-that-require-more-privileges-than-default

Ref https://github.com/SUSE/fissile/wiki/Operator-overrides-of-role-capabilities

SCF test PR https://github.com/SUSE/scf/pull/1511


kube/pod.go
```diff
@@ -450,8 +459,36 @@ func getSecurityContext(role *model.Role) helm.Node {
...
 		return nil
 	}
+	if createHelmChart {
+		// Conditional capabilities, fixed set, and range over
+		// operator-specified dynamic set.
+		caplist := helm.NewList()
+		for _, cap := range capabilities {
+			caplist.Add(cap)
+		}
+		caplist.Add(helm.NewNode("{{ . }}", helm.Block(fmt.Sprintf("range %s", config))))
```
mook-as: Can we `|upper` this, so that it isn't case-sensitive for the operator?

kube/pod.go
```diff
@@ -450,8 +459,36 @@ func getSecurityContext(role *model.Role) helm.Node {
 		capabilities = append(capabilities, cap)
 	}
 	if len(capabilities) == 0 {
+		if createHelmChart {
```
mook-as: The `len(capabilities) == 0` chunk is basically only relevant if 1) we're _not_ building helm charts, and 2) there are no capabilities. We don't need to make the helm chart path complicated like this, just `range` over an empty / `nil` list instead.

kube/pod_test.go
```diff
@@ -2047,28 +2066,191 @@ func TestPodCPUAHelmActive(t *testing.T) {
...
+			}
+			actual, err := testhelpers.RoundtripNode(sc, config)
+			if !assert.NoError(err) {
+				return
+			}
+			testhelpers.IsYAMLEqualString(assert, `---
+				capabilities:
+					add:
+					-	"SOMETHING"
+					-	"something"
```
mook-as: I wonder if it would make more sense to use `assert.Contains` instead (walking down the struct instead of comparing YAMLs)

kube/pod_test.go
```diff
@@ -2047,28 +2066,191 @@ func TestPodCPUAHelmActive(t *testing.T) {
...
+	assert := assert.New(t)
+
+	role := podTemplateTestLoadRole(assert)
+	if role == nil {
 		return
 	}
 
-	actual, err := testhelpers.RoundtripKube(sc)
-	if !assert.NoError(err) {
+	// Clear the capability list from the manifest to force fully-privileged mode.
```
mook-as: Umm, doesn't that force _minimally_ privileged mode instead?

model/roles.go
```diff
@@ -1501,6 +1501,17 @@ func (r *Role) LookupJob(name string) *RoleJob {
 	return nil
 }
 
+// IsPrivileged tests if the role capabilities enable fully privileged
+// mode.
+func (r *Role) IsPrivileged() bool {
+	for _, cap := range r.Run.Capabilities {
```
mook-as: I think this needs to be always upper case (regardless of which capability)? I believe we have some input normalization going on when we read in the role manifest; maybe upper case all the capabilities there, instead of doing this here (potentially repeatedly)?

kube/pod.go
andreas-kupries: Ah. Certainly.

andreas-kupries: Done with commit 6d720721499381c95c34e4e4c9960da8bb23e60a

kube/pod.go
andreas-kupries: Ok, we have lots of roles with (2). Pretty much all but one.
However I am not sure about (1).  When I am building helm charts I __do__ hit this new code for all these roles with (2). So it looks to be needed.

kube/pod_test.go
andreas-kupries: While walking the structure directly might be more efficient I like the readability of the current code.
It feels to be (very) near to go's example-based testing facilities.

kube/pod_test.go
andreas-kupries: Right. And actually `default privileges`, might not be minimal. Will fix.

andreas-kupries: Done with commit b5563f564ddf71072c25c524d9d4c29e43733e18

kube/pod.go
mook-as: Yes, currently it would hit this code; I'm saying we can change this to:
```golang
if createHelmChart {
	caplist := helm.NewList()
	for _, cap := range capabilities {
		// If there are no capabilities, this just never gets added
		caplist.Add(cap)
	}
	caplist.Add(helm.NewNode("{{ . | upper }}", helm.Block(fmt.Sprintf("range %s", config))))
	cla := helm.NewMapping("add", caplist)
	cla.Set(helm.Block(notAll))
	sc := helm.NewMapping()
	sc.Add("privileged", helm.NewNode(true, helm.Block(hasAll)))
	sc.Add("capabilities", cla)
	return sc
}
if len(capabilities) == 0 {
	return nil
}
return helm.NewMapping("capabilities", helm.NewMapping("add", helm.NewNode(capabilities)))
```
}

model/roles.go
andreas-kupries: In the next line it uses `ToUpper` to normalize for the comparison.
The input normalization you think of is likely the `getSecurityContext` in pod.go.
See https://github.com/SUSE/fissile/pull/351/files/7e2a6f97e5369d540dc4db413e5f69473260cc4e#diff-a563c88f5954ff16e94d3ec81b1de52cR454

`IsPrivileged` is a variant of that which only does the checking for ALL without saving the transformed capabilities as `gSC` does.

Oh. You are asking to move the normalization from `IsP` / `gSC` to the where the manifest is read.
Should be possible. Will try to find where that happens.
Likely in the various validations.

kube/pod.go
andreas-kupries: I believe I see now. Move the block dealing with mode `caplist` to the front. That way it can also deal with mode `nil` also, and the entire specialization for `nil` can be removed. It makes a small difference in that mode `nil` now also uses a range instead of direct assignment of the user's cap-list. Bearable.

Will do.

And sorry, while with re-reading your first comment I am now seeing a glimmer of how you implied this optimization this is definitely hindsight.

mook-as: Sorry, I should have explained it better originally; ended up falling back to code because, that, unlike English, is unambiguous (hopefully). :grin:

model/roles.go
andreas-kupries: Done with commit ef1155c4a44ae9f883dfa6dddfd16cfa09a18d91

andreas-kupries:
Commit a210aceb5826a0d4aaeb7d0e5a41aec588b80148 provides user documentation for the new feature.

andreas-kupries:
Commit 97bf4ab5299f1c97d0cae6c3ccd3b3bf43123cb5 does more `t.Parallel()`.

kube/job_test.go
```diff
@@ -73,6 +74,7 @@ func TestJobPreFlight(t *testing.T) {
 }
 
 func TestJobPostFlight(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
andreas-kupries: You are welcome.

kube/pod.go
```diff
@@ -440,15 +440,43 @@ func getEnvVarsFromConfigs(configs model.ConfigurationVariableSlice, settings Ex
...
 }
 
-func getSecurityContext(role *model.Role) helm.Node {
+func getSecurityContext(role *model.Role, createHelmChart bool) helm.Node {
+	var hasAll string
+	var notAll string
+	var config string
+	if createHelmChart {
+		config = fmt.Sprintf(".Values.sizing.%s.capabilities", makeVarName(role.Name))
+		hasAll = fmt.Sprintf("if has \"ALL\" %s", config)
```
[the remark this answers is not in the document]
andreas-kupries: Sure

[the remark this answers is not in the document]
andreas-kupries: Done with commit 2e8af97c37a2eb2dfbeb99591cdf23a5303f6b20
Plus indent fix.

PR #352, opened 2018-05-07
Add support for multi-containers per pod

### LINKED ISSUE #340: Add support for multi-containers aka sidecar per Pod

HeavyWombat:
Like discussed in #340 and in a call with @viovanov, we would like to propose the following:
- Add a new `type` of role called `colocated-container` which is only intended to be used by another role and will not end up as a template YAML itself.
- Add a new list of role names called `colocated_container` in the role, which contains the name references to `colocated-container` roles.

A series of validation tests were added:
- Check for unused `colocated-container` roles
- Check for `colocated-container` not existing
- Check if a role and its `colocated-container` roles have port collisions

**One important point** about `monit`: In this proposal we avoided adding use case specific logic for `monit` into `fissile`. Each `colocated-container` role using `monit` has to specify a `monit` port that is sure not to collide with the main role or any other `colocated-container` of that main role. We felt this is better suited in the README about this feature. If this is not a desired behavior, please let us know.

We would be happy to provide further documentation on that feature, if you could point us to the best spot for that.

This would fix #340.

Signed-off-by: Enrique Eduardo Encalada Olivas <encalada@de.ibm.com>

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -1079,7 +1079,7 @@ func (f *Fissile) roleHasStorage(role *model.Role) bool {
 
 func (f *Fissile) generateKubeRoles(settings kube.ExportSettings) error {
 	for _, role := range settings.RoleManifest.Roles {
-		if role.IsDevRole() {
+		if role.IsDevRole() || role.IsColocatedContainerRole() {
 			continue
 		}
 		if settings.CreateHelmChart && role.Run.FlightStage == model.FlightStageManual {
--- a/kube/pod.go
+++ b/kube/pod.go
@@ -26,102 +26,20 @@ func NewPodTemplate(role *model.Role, settings ExportSettings, grapher util.Mode
 		return nil, fmt.Errorf("Role %s has no run information", role.Name)
 	}
 
-	roleName := strings.Replace(strings.ToLower(role.Name), "_", "-", -1)
-	roleVarName := makeVarName(roleName)
-
-	vars, err := getEnvVars(role, settings)
-	if err != nil {
-		return nil, err
-	}
-
-	var resources helm.Node
-	var requests *helm.Mapping
-	var limits *helm.Mapping
-
-	if settings.UseMemoryLimits || settings.UseCPULimits {
-		requests = helm.NewMapping()
-		limits = helm.NewMapping()
-		resources = helm.NewMapping("requests", requests, "limits", limits)
-	}
-
-	if settings.UseMemoryLimits {
-		if settings.CreateHelmChart {
-			requests.Add("memory",
-				helm.NewNode(fmt.Sprintf("{{ int .Values.sizing.%s.memory.request }}Mi", roleVarName),
-					helm.Block(fmt.Sprintf("if and .Values.sizing.memory.requests .Values.sizing.%s.memory.request", roleVarName))))
-			limits.Add("memory",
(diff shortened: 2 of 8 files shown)


viovanov:
I think fissile should validate that jobs that use `colocated_containers` reference the necessary shared volumes.

viovanov:
Update: still trying to run a `metron` side-container.
I've got a pod containing the `diego-cell` and `metron` containers.
The `diego-cell` starts up, then just stops and restarts.

HeavyWombat:
Commited https://github.com/SUSE/fissile/pull/352/commits/d0b9a77d17c62a064b9e478f80cc323f6744847b which includes the promised missing validation of the shared volumes. Please let me know what you think. Thanks in advance.

alexjh:
@HeavyWombat Can you look at the conflict in `kube/pod.go`? I'm hoping we can merge this soon as we have other fissile branches coming in that will probably create even more conflicts. :)

viovanov:
I've approved the changes - once the conflict is resolved, we can merge. Thanks!

HeavyWombat:
I'll have a look at the merge conflict as soon as possible.

PR #354, opened 2018-05-24
Rework pod management

mook-as:
This implements the [Pod Management using Role Manifest Tags](https://github.com/SUSE/fissile/wiki/Pod-Management-using-Role-Manifest-Tags) document.

Notable changes:
- We only use statefulset now
- The tags are all changed
- Dev-only roles are no longer a thing (it really hasn't for a while, since we deploy via kube now)
- BOSH roles will actually look at `monit` for the readiness
- For active/passive roles, a command line is used to determine when it's ready (at which point a label is applied to the pod)
- I use lists of constants (since we don't have actual enums in golang) more, and raw strings less.
- For some reason, the old kube stuff was (incorrectly) outputting `externalIPs` as a single string instead of a list

This should probably wait for #352 to land first — that started first, and making that rebase sounds annoying.

/cc @HeavyWombat @zhangtbj as a heads-up.

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -1079,7 +1079,7 @@ func (f *Fissile) roleHasStorage(role *model.Role) bool {
 
 func (f *Fissile) generateKubeRoles(settings kube.ExportSettings) error {
 	for _, role := range settings.RoleManifest.Roles {
-		if role.IsDevRole() || role.IsColocatedContainerRole() {
+		if role.IsColocatedContainerRole() {
 			continue
 		}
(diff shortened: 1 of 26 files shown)


viovanov:
@mook-as can you please resolve the conflict? why is that line commented out?

alexjh:
Since Vlad and I have both been confused by it, I think there should be a comment on that commented out code in `helm_helper.go`.

mook-as:
FWIW, it was commented out because there's a lot of false positives (where we test that failing to execute the template returns an error), which means lots of useless spew.  That was especially confusing as stdout/stderr are not synchronized.

PR #355, opened 2018-05-24
Add HELM_IS_INSTALL built-in variable

jandubois:
It will be used by the secrets-generator to determine if previous secrets should exist or not.

### DIFF
--- a/kube/pod.go
+++ b/kube/pod.go
@@ -398,6 +398,15 @@ func getEnvVarsFromConfigs(configs model.ConfigurationVariableSlice, settings Ex
 			continue
 		}
 
+		if config.Name == "HELM_IS_INSTALL" {
+			value := "true"
+			if settings.CreateHelmChart {
+				value = "{{ .Release.IsInstall | quote }}"
+			}
+			env = append(env, helm.NewMapping("name", config.Name, "value", value))
+			continue
+		}
+
 		if config.Name == "KUBE_SECRETS_GENERATION_COUNTER" {
 			value := "1"
 			if settings.CreateHelmChart {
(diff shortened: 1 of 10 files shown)


kube/pod.go
```diff
@@ -370,6 +370,15 @@ func getEnvVarsFromConfigs(configs model.ConfigurationVariableSlice, settings Ex
 			continue
 		}
 
+		if config.Name == "HELM_IS_INSTALL" {
+			value := "true"
```
mook-as: That's the string `true`, not the boolean value `true`, right? That's… actually probably the right thing to do in an environment variable, I guess.

kube/pod.go
```diff
@@ -370,6 +370,15 @@ func getEnvVarsFromConfigs(configs model.ConfigurationVariableSlice, settings Ex
...
+			if settings.CreateHelmChart {
+				value = "{{ .Release.IsInstall | quote }}"
```
mook-as: Can this be unset if false? So that `test -n "${HELM_IS_INSTALL}"` would do the right thing.

kube/pod.go
jandubois: Yes, env vars are always strings.

kube/pod.go
jandubois: It will be either "true" or "false", which works fine if you want to pass it as a `flag.Bool` to a Go binary (which I do). Using an empty string would work badly, so I would prefer to keep it as it is.

kube/pod.go
jandubois: Now the real question is: should this be `true` or `false` in a kube config? I guess it doesn't really matter, as it should not be used by the bosh-tasks. So maybe it would be better not to set it at all in that case. Not sure any of this matters; I hope the pure kube config files go away at some point...

mook-as: I'm not too worried about the kube versions; they're already busted  (`external_ips` is a string instead of a list), and nobody complained. (I'm fixing it in #354 anyway).

Whatever value you put here is fine; even `unknown` or `pants`.

PR #356, opened 2018-05-25
Add configurable annotations to volumes

qu1queee:
Hi, 

This PR contains a new field in the `RoleRunVolume` structure,  to allow persistent and shared
volumes annotations to be customizable.  This is for use cases where the default annotation is not enough for the IaaS, and additional ones are required. Also, the default one, should be allow to be overwritten based on the user needs.

### DIFF
--- a/kube/stateful_set.go
+++ b/kube/stateful_set.go
@@ -68,11 +68,12 @@ func getVolumeClaims(role *model.Role, createHelmChart bool) []helm.Node {
 		}
 
 		meta := helm.NewMapping("name", volume.Tag)
-		if volume.Annotations == nil || len(volume.Annotations) == 0 { // default behaviour
-			meta.Add("annotations", helm.NewMapping(VolumeStorageClassAnnotation, storageClass))
-		} else {
-			meta.Add("annotations", volume.Annotations)
+		annotationList := helm.NewMapping()
+		annotationList.Add(VolumeStorageClassAnnotation, storageClass)
+		for key, value := range volume.Annotations {
+			annotationList.Add(key, value)
 		}
+		meta.Add("annotations", annotationList)
 
 		var size string
 		if createHelmChart {
--- a/kube/stateful_set.go
+++ b/kube/stateful_set.go
@@ -68,7 +68,11 @@ func getVolumeClaims(role *model.Role, createHelmChart bool) []helm.Node {
 		}
 
 		meta := helm.NewMapping("name", volume.Tag)
-		meta.Add("annotations", helm.NewMapping(VolumeStorageClassAnnotation, storageClass))
+		if volume.Annotations == nil || len(volume.Annotations) == 0 { // default behaviour
+			meta.Add("annotations", helm.NewMapping(VolumeStorageClassAnnotation, storageClass))
+		} else {
+			meta.Add("annotations", volume.Annotations)
+		}
 
 		var size string
 		if createHelmChart {
--- a/test-assets/role-manifests/kube/volumes-with-annotation.yml
+++ b/test-assets/role-manifests/kube/volumes-with-annotation.yml
@@ -10,26 +10,20 @@ roles:
     scaling:
       min: 1
       max: 2
-    persistent-volumes:
+    volumes:
     - path: /mnt/persistent
+      type: persistent
       tag: persistent-volume
       size: 5 # parsecs
       annotations:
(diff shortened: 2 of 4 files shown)


test-assets/role-manifests/kube/volumes-with-annotation.yml
```diff
@@ -0,0 +1,35 @@
...
+  jobs:
+  - name: tor
+    release_name: tor
+  run:
+    capabilities:
+    - something
+    scaling:
+      min: 1
+      max: 2
+    persistent-volumes:
```
mook-as: This should probably be gradually moved to:
```yaml
volumes:
- path: /mnt/persistent
  type: persistent
  …
- path: /mnt/shared
  type: shared
  …
- path: /sys/fs/cgroup
  type: host
  …
```

test-assets/role-manifests/kube/volumes-with-annotation.yml
```diff
@@ -0,0 +1,35 @@
...
+    shared-volumes:
+    - path: /mnt/shared
+      tag: shared-volume
+      size: 40 # cakes
+      annotations:
+    volumes:
+    - path: /sys/fs/cgroup
+      type: host
+      tag: host-volume
+configuration:
```
mook-as: We probably don't need the configuration stuff in this test? I find that keeping the test files limited to what we're testing makes it easier to understand.  YMMV.

test-assets/role-manifests/kube/volumes-with-annotation.yml
qu1queee: I dont follow you here.

qu1queee:
@mook-as I addressed your comments, and push another commit(did not wanted to amend it). Pls, take a look.

kube/stateful_set.go
```diff
@@ -68,7 +68,14 @@ func getVolumeClaims(role *model.Role, createHelmChart bool) []helm.Node {
 		}
 
 		meta := helm.NewMapping("name", volume.Tag)
-		meta.Add("annotations", helm.NewMapping(VolumeStorageClassAnnotation, storageClass))
+		annotationList := helm.NewMapping()
+		annotationList.Add(VolumeStorageClassAnnotation, storageClass)
+		if volume.Annotations != nil || len(volume.Annotations) > 0 {
```
mook-as: I don't think you need the `if` anymore, since (IIRC) `range nil {…}` just runs zero iterations. Even if it didn't, you don't need the `volume.Annotations != nil` (because `len(nil)` is zero).  And even if needed _that_, you wouldn't want an `||`, but a `&&` :)

test-assets/role-manifests/kube/volumes-with-annotation.yml
mook-as: We're basically keeping the `persistent-volumes` and `shared-volumes` keys for backwards compatibility; we're trying to move towards `volumes` with individual types to have more uniform support for `host` and `empty` volume mount types.  So having the backwards compatible forms in the test cases doesn't seem to make much sense.

kube/stateful_set.go
qu1queee: terrible that `||`, thanks for spotting that mistake.

test-assets/role-manifests/kube/volumes-with-annotation.yml
qu1queee: ok, I got it. I think I start the PR based on the [postgres role](https://github.com/SUSE/scf/blob/develop/container-host-files/etc/scf/config/role-manifest.yml#L386)  and that did not help in considering the new `volumes` setup. Any reasons why postgres still using `persistent-volumes` key?

qu1queee:
@mook-as I just amend my changes(avoiding many commits), pls take a look.

kube/stateful_set.go
```diff
@@ -68,7 +68,12 @@ func getVolumeClaims(role *model.Role, createHelmChart bool) []helm.Node {
 		}
 
 		meta := helm.NewMapping("name", volume.Tag)
-		meta.Add("annotations", helm.NewMapping(VolumeStorageClassAnnotation, storageClass))
+		annotationList := helm.NewMapping()
+		annotationList.Add(VolumeStorageClassAnnotation, storageClass)
```
mook-as: (Technically, this can be `annotationList := helm.NewMapping(VolumeStorageClassAnnotation, storageClass)` but that really isn't worth your time or mine)

PR #357, opened 2018-05-28
Move non-role settings out of sizing to config

andreas-kupries:
Ref: https://trello.com/c/7xGqgWkI/671-3-make-sizing-section-of-valuesyaml-uniform

Moved non-role sizing keys to config.
    
- Modified output for values.yaml
- Updated references in the generated templates
- Added guards against the old keys to the generated templates.
- Updated existing tests
- Added documentation about the sizing cleanup.
- Added tests triggering the new guards against old variable names.
- Added more tests for the replica guards.

:warning: Disabled parallel execution. Ran into data race deep in the template engine when having more than 2 to 3 parallel sub tests in the deployment tests. Render tweak, new variable for result instead of overwriting the argument (did not help with the race).

### DIFF
--- a/docs/sizing-cleanup.md
+++ b/docs/sizing-cleanup.md
@@ -1,0 +1,48 @@
+# Issue
+
+Parsing the `sizing` section of `values.yaml` is difficult because it
+contains a mix of per-role descriptions and other things: HA, memory,
+cpu.
+
+The extra tuneables should be moved somewhere else.
+
+Example of an old variable setting: `--set sizing.HA=true`
+
+# Implemented change
+
+Moved the non-role-specific settings to the new key `config`.
+
+For example `sizing.HA` becomes `config.HA`.
+
+The keys affected by this change are:
+
+   * `sizing.HA`
+   * `sizing.cpu.limits`
+   * `sizing.cpu.requests`
+   * `sizing.memory.limits`
+   * `sizing.memory.requests`
+
+The only keys left under `sizing` are the per-role descriptions.
+
+Further, to prevent users from accidentally using the old names in
+their overide yaml files, all templates are extended to contain
+guarding statements of the form
+
+```
+    {{- if .Values.FOO }}
+    _moved(FOO): {{ fail "Bad use of moved variable FOO. The new name to use is [FOO]" }}
+    {{- end }}
+```
+
+where `FOO` is one of the keys above, `(FOO)` that key changed to not
+be nested (`.` --> `_`), and `[FOO]` the new key for `FOO`.
+
+Example: `sizing.cpu.limits` to `sizing_cpu_limits`
+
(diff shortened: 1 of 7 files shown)


kube/deployment_test.go
```diff
@@ -95,16 +95,101 @@ func TestNewDeploymentHelm(t *testing.T) {
...
+		// t.Parallel()
+		config := map[string]interface{}{
+			"Values.sizing.role.count":                 "0",
+			"Values.sizing.role.affinity.nodeAffinity": "snafu",
+			"Values.sizing.role.capabilities":          []interface{}{},
+			"Values.kube.registry.hostname":            "docker.suse.fake",
+			"Values.kube.organization":                 "splat",
+			"Values.env.KUBE_SERVICE_DOMAIN_SUFFIX":    "domestic",
+		}
+		_, err = testhelpers.RenderNode(deployment, config)
```
mook-as: Your data race is because the same `err` is being assigned to

testhelpers/helm_helper.go
```diff
@@ -147,6 +147,6 @@ func renderInclude(name string, data interface{}) (string, error) {
 	// require adding the handling of `associated` templates.  A
 	// first run at this generated a stack overflow.  The fake
 	// simply shows what path/name would have been included.
-	name = filepath.Base(name)
-	return name, nil
+	base := filepath.Base(name)
+	return base, nil
```
mook-as: If you're going to change this anyway, it might as well just be `return filepath.Base(name), nil`?

docs/sizing-cleanup.md
```diff
@@ -0,0 +1,48 @@
+# Issue
```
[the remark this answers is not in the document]
andreas-kupries: The following description is of the state. Still, I guess I can remove the intro section.

kube/deployment_test.go
andreas-kupries: Hm. Maybe. I will definitely fix that and hope for the best. However the stack traces go gave me show something deep in the template render code, not the assignment to err. Well, we will see.

testhelpers/helm_helper.go
andreas-kupries: True. Will change.

Note, general, changes as I find time in the 1.36 bump.

andreas-kupries: Done with commit 002f7286fb9344c902d70202da9a5a2f56ea39d0

andreas-kupries:
All comments addressed.

kube/deployment_test.go
mook-as: I believe the issue was that we were printing out giant YAML things when the templates failed to expand or something, on `stdout`, which was between lines of the stack on `stderr`, or something along those lines.

PR #359, opened 2018-05-29
Only use external ips for non loadbalanced mode

alexjh:
Automatically provisioning load balancers in Azure (and likely other container services) was broken because of two issues:

* load balancers were being created for non-public services
* external IPs were being defined for public services even though a load balancer was requested

These changes only use `externalIP`s if `loadbalanced == False`, and only request load balancers for `-public` services. A public service is created if the role manifest's `exposed-port` definition has `public: true`

In the second commit I simplified the `ClusterIP` setting since it's the default. It wraps the loadbalancer type in a block rather than the if/else statement. I feel that's simpler, although it's less obvious for the `ClusterIP` setting because that's not explicit any more. If there's any objections I can remove that commit from the PR because it doesn't change functionality.

### DIFF
--- a/kube/service.go
+++ b/kube/service.go
@@ -146,16 +146,13 @@ func newService(role *model.Role, serviceType newServiceType, settings ExportSet
 	}
 	spec.Add("selector", selector)
 
-	// All services should use ClusterIP except the public one which is overwritten below
-	spec.Add("type", "ClusterIP")
-
 	if serviceType == newServiceTypeHeadless {
 		spec.Add("clusterIP", "None")
 	}
 	if serviceType == newServiceTypePublic {
 		if settings.CreateHelmChart {
 			spec.Add("externalIPs", "{{ .Values.kube.external_ips | toJson }}", helm.Block("if not .Values.services.loadbalanced"))
-			spec.Add("type", "{{ if .Values.services.loadbalanced }} LoadBalancer {{ else }} ClusterIP {{ end }}")
+			spec.Add("type", "LoadBalancer", helm.Block("if .Values.services.loadbalanced"))
 		} else {
 			spec.Add("externalIPs", []string{"192.168.77.77"})
 		}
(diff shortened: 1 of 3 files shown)


kube/service.go
```diff
@@ -138,11 +133,17 @@ func NewClusterIPService(role *model.Role, headless bool, public bool, settings
...
 	}
 	if public {
-		externalIPs := "[ 192.168.77.77 ]"
 		if settings.CreateHelmChart {
-			externalIPs = "{{ .Values.kube.external_ips | toJson }}"
+			externalIPs := "{{ .Values.kube.external_ips | toJson }}"
+			spec.Add("externalIPs", externalIPs, helm.Block("if not .Values.services.loadbalanced"))
+			spec.Add("type", "{{ if .Values.services.loadbalanced }} LoadBalancer {{ else }} ClusterIP {{ end }}")
+		} else {
+			externalIPs := "[ 192.168.77.77 ]"
```
mook-as: This should be `[]string{"192.168.77.77"}` (the old code was wrong); I actually fixed it in #354 but that hasn't landed yet so we're in a race :grin:

alexjh: Ok, I'm going to take #354 for review because it looks like we changed things in the same area.

alexjh:
@viovanov This PR fixes the load balancer problem. I'm rebasing on Mark's active/passive changes and hopefully it will be more obvious at that point. I'll update the description here too though.

viovanov:
I still don't understand how this fixes it.
Can you please point me to the ticket for this?

alexjh:
https://trello.com/c/sQGhnAwJ/670-make-the-azure-load-balancer-work

alexjh:
@viovanov I've rebased and pushed my branch based on Mark's changes. The actual commit is ca412fd8aa2a773d7ee25ea929d62f4c92835315.

alexjh:
Rebased on master now that the active/passive work landed.

mook-as:
I'm going to dismiss Vlad's review based on the code (as it is) looks clear enough. I suspect it wasn't at the point the comment was made.  The old commits are gone, though, so I can't just look at those either…

alexjh:
Cool, that's even better since it's been tested on ACS **and** AKS now.

PR #360, opened 2018-05-30
Fix the generated guard conditions.

andreas-kupries:
Cannot check FOO.BAR when FOO is nil.
Cannot `and FOO FOO.BAR` because `and` is not short-circuited.
Matched tests (changed column indices in the error messages)

Ref https://trello.com/c/7xGqgWkI/671-3-make-sizing-section-of-valuesyaml-uniform

### DIFF
--- a/kube/deployment.go
+++ b/kube/deployment.go
@@ -104,21 +104,13 @@ func generalCheck(role *model.Role, controller *helm.Mapping, settings ExportSet
 	// (2) The `and` operator is not short cuircuited, it evals all of its arguments
 	// Thus `and FOO FOO.BAR` will not work either
 
+	fail := `{{ fail "Bad use of moved variable sizing.HA. The new name to use is config.HA" }}`
+	controller.Add("_moved_sizing_HA", fail, helm.Block("if .Values.sizing.HA"))
+
 	for _, key := range []string{
-		"HA",
 		"cpu",
 		"memory",
 	} {
-		if key == "HA" {
-			guardVariable := fmt.Sprintf("_moved_sizing_%s", key)
-			block := fmt.Sprintf("if .Values.sizing.%s", key)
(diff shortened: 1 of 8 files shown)


kube/deployment.go
```diff
@@ -99,20 +98,41 @@ func generalCheck(role *model.Role, controller *helm.Mapping, settings ExportSet
...
-			key, key)
+		if key == "HA" {
+			guardVariable := fmt.Sprintf("_moved_sizing_%s", key)
+			block := fmt.Sprintf("if .Values.sizing.%s", key)
+			fail := fmt.Sprintf(`{{ fail "Bad use of moved variable sizing.%s. The new name to use is config.%s" }}`,
+				key, key)
+
+			controller.Add(guardVariable, fail, helm.Block(block))
+			continue
+		}
```
jandubois: Not sure why you put this block inside the loop, and then just filter it out. Just move it on top of the loop, and get rid of the conditional. Also saves the `fmt.Sprintf` to setup `fail` because the `key` is already known.

There will never be any additional entries for this loop, as this is a one-time migration, and there are no other non-rolename keys left in `.Values.sizing`, so no reason to not make this code as simple as possible.

jandubois:
```
			fail := `{{ fail "Bad use of moved variable sizing.HA. The new name to use is config.HA" }}`
			controller.Add("_moved_sizing_HA", fail, helm.Block("if .Values.sizing.HA"))
```

andreas-kupries: Somehow my mind was wedded to the loop. Agreed, specialized this one case is more sensible.
To come.

andreas-kupries: Done with commit 6a2949581a367e0a8e743750e1dca34f32446451

andreas-kupries:
Thinking about it, the test at line 191 ought to have caught this. And I now wonder why it did not.

PR #363, opened 2018-06-25
Add support for structured properties

jandubois:


jandubois:
@mook-as Sorry I asked for review already; just noticed that `example` values in bosh specs are multi-line strings and not actual structured values, so will adopt to stay compatible.

Hence `WIP` label for now.

jandubois:
https://github.com/SUSE/fissile/pull/363/commits/9294da15873aa23d1f328005d6ed771e119a0df1 is no longer required for this PR, but I would prefer to keep it because it is working, tested code that we may want to use in the future.

helm/config.go
```diff
@@ -441,16 +453,23 @@ func NewEncoder(writer io.Writer, modifiers ...func(*Encoder)) *Encoder {
...
 	}
 	enc.Set(modifiers...)
 	return enc
 }
 
-// Encode writes the config mapping held by the node to the stream.
-func (enc *Encoder) Encode(node Node) error {
+// Encode writes the config mapping held by the node to the stream. The optional
+// argument can be used to specify a prefix to indent the whole generated document.
+func (enc *Encoder) Encode(node Node, optional ...string) error {
```
mook-as: Not a fan of optional arguments like this; it's hard to figure out what optional parameters are, and doesn't work if we happen to want another non-string one.  Of course, I don't know of any good way to do that in golang either… Making the prefix non-optional might be the best we have.

helm/config.go
```diff
@@ -510,22 +529,28 @@ func useOnce(prefix *string) string {
 // each line will include at least a single word, even if it exceeds the
 // wrapping column).
 //
-// Paragraphs starting with "* " or "- " are treated as bullet points, so
-// wrapped lines will be indented by 2 spaces.
+// Paragraphs starting with "*" or "-", followed by one or more spaces, are treated as
+// bullet points, so wrapped lines will be indented to the same level.
+//
+// Similary any paragraphs intented by a string of spaces will keep this indent for
```
jandubois: indented

helm/config.go
jandubois: I'm fine with removing this; it is not actually being used after all.

If we ever *do* want this feature, then we could turn this into an additional option on the encoder object itself. We already have  `Indent()` for the per-level indentation, so maybe `InitialIndent()` or whatever. Not going to implement it now anyways; but I can remove the optional parameter.

PR #365, opened 2018-07-13
Fix port collition check problem

### LINKED ISSUE #364: Colocated role port collision check fails for same port with TCP/UDP
With PR #340, a collision check for colocated roles has been introduced that has an issue when you specify the same port number with different protocols, because the collision check only checks the port number.

```yaml
- name: consul
  colocated_containers:
  - logs-forwarder
  run:
    exposed-ports:
    - name: serf-tcp
      extenral: 8301
      internal: 8301
      protocol: TCP
    - name: serf-udp
      extenral: 8301
      internal: 8301
      protocol: UDP
```

The configuration is totally fine, but the collision checks reports that as an issue: `Error loading roles manifest: role[consul]: Invalid value: 8301: port collision, the same port is used by: consul, consul`

To resolve this, the combination of protocol and port has to be checked against one another, for example: `TCP-8301` and `UDP-8301`. I will start to work on a fix for this as soon as possible.

HeavyWombat:
Replace lookup logic used in port collision check to use the tuple of
protocol and port (`TCP/53`). That way a port collision is only counted
in case both roles use the same protocol. The same port used by both
protocols is no longer falsely reported as a configuration issue.

This should fix #364.

### DIFF
--- a/model/roles.go
+++ b/model/roles.go
@@ -1632,40 +1632,40 @@ func validateColocatedContainerPortCollisions(RoleManifest *RoleManifest) valida
 
 	for _, role := range RoleManifest.Roles {
 		if len(role.ColocatedContainers) > 0 {
-			portsMap := map[int][]string{}
+			lookupMap := map[string][]string{}
 
 			// Iterate over this role and all colocated container roles and store the
-			// names of the roles for each port (there should be no list with more
-			// than one entry)
+			// names of the roles for each protocol/port (there should be no list with
+			// more than one entry)
 			for _, toBeChecked := range append([]*Role{role}, role.GetColocatedRoles()...) {
 				for _, exposedPort := range toBeChecked.Run.ExposedPorts {
 					for i := 0; i < exposedPort.Count; i++ {
-						port := exposedPort.ExternalPort + i
-						if _, ok := portsMap[port]; !ok {
-							portsMap[port] = []string{}
+						protocolPortTuple := fmt.Sprintf("%s/%d", exposedPort.Protocol, exposedPort.ExternalPort+i)
+						if _, ok := lookupMap[protocolPortTuple]; !ok {
+							lookupMap[protocolPortTuple] = []string{}
 						}
 
-						portsMap[port] = append(portsMap[port], toBeChecked.Name)
+						lookupMap[protocolPortTuple] = append(lookupMap[protocolPortTuple], toBeChecked.Name)
 					}
 				}
 			}
 
-			// Get a sorted list of the keys (port)
-			ports := []int{}
-			for port := range portsMap {
-				ports = append(ports, port)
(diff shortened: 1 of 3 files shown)


model/roles.go
```diff
@@ -1632,40 +1632,40 @@ func validateColocatedContainerPortCollisions(RoleManifest *RoleManifest) valida
...
 	for _, role := range RoleManifest.Roles {
 		if len(role.ColocatedContainers) > 0 {
-			portsMap := map[int][]string{}
+			lookupMap := map[string][]string{}
 
 			// Iterate over this role and all colocated container roles and store the
-			// names of the roles for each port (there should be no list with more
-			// than one entry)
+			// names of the roles for each protocol/port (there should be no list with
+			// more than one entry)
```
mook-as: This works because we validate the protocol name, case-sensitively, before this happens, right?

model/roles_test.go
```diff
@@ -1232,8 +1232,28 @@ func TestLoadRoleManifestColocatedContainersValidationPortCollisions(t *testing.
 	roleManifestPath := filepath.Join(workDir, "../test-assets/role-manifests/model/colocated-containers-with-port-collision.yml")
 	roleManifest, err := LoadRoleManifest(roleManifestPath, []*Release{torRelease, ntpRelease}, nil)
 	assert.Nil(roleManifest)
-	assert.EqualError(err, "role[main-role]: Invalid value: 80: port collision, the same port is used by: main-role, to-be-colocated\n"+
-		"role[main-role]: Invalid value: 10443: port collision, the same port is used by: main-role, to-be-colocated")
+	assert.EqualError(err, "role[main-role]: Invalid value: \"TCP/10443\": port collision, the same protocol/port is used by: main-role, to-be-colocated"+"\n"+
+		"role[main-role]: Invalid value: \"TCP/80\": port collision, the same protocol/port is used by: main-role, to-be-colocated")
```
mook-as: Ah, the sorting is off because it's alpha-sort, not numeric. That should be okay :D

model/roles.go
HeavyWombat: Yes, exactly. Now that you mention it, I forgot to point that out. I checked that only `TCP` or `UDP` (uppercase) are possible protocol content and therefore assume it will be fine to concatenate that with a slash to the port number.

model/roles_test.go
HeavyWombat: That swap annoyed me a little bit. But on the other side, keeping the numeric sorting would be possible, but would require a Golang version 1.8 feature to sort with a custom sorting function. Since this is only for the error use case, I figured one could live with that missing numeric sort order. Let me know, if you would like it any other way.

HeavyWombat:
Thanks for looking into that so quickly!

PR #375, opened 2018-07-26
Don't use "monit summary" in container readiness probe

viovanov:
On my deployments I see monit using a lot of CPU because of this.
Replacing with a curl call that grabs status from the monit daemon using using the HTTP api.

### DIFF
--- a/scripts/dockerfiles/readiness-probe.sh
+++ b/scripts/dockerfiles/readiness-probe.sh
@@ -17,7 +17,7 @@
 set -o errexit -o nounset
 
 # Grab monit port
-monit_port=$(cat /etc/monitrc | grep "httpd port" | awk '{print $4}')
+monit_port=$(awk '/httpd port/ { print $4 }' /etc/monitrc)
 
 # Check that monit thinks everything is ready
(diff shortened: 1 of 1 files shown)


viovanov:
This improved performance in my local deployments.

scripts/dockerfiles/readiness-probe.sh
```diff
@@ -16,12 +16,14 @@
 
 set -o errexit -o nounset
 
+# Grab monit port
+monit_port=$(cat /etc/monitrc | grep "httpd port" | awk '{print $4}')
```
mook-as: That's the same as `awk '/httpd port/ { print $4 }' /etc/monitrc`?

andreas-kupries: I believe so, yes. Guilty of similar constructions myself. Why use one process if you can have three ;)

aduffeck: @mook-as @viovanov asked me to handle this since he's on the road.

I verified that your suggestion works and changed the code accordingly.

mook-as:
Thanks! Merged.

Issue #370, opened 2018-07-22
Make the role manifest compatible with BOSH deployment manifests


viovanov:
- [x] Rename `roles` to `instance_groups`
- [x] For jobs definitions, use `release` instead of `release_name`
- [x] Change secret generator so it supports consuming information from a [`variables`](https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml#L1249) section
- [x] Convert variables and secret information to a [`variables`](https://github.com/cloudfoundry/cf-deployment/blob/master/cf-deployment.yml#L1249) section in SCF/UAA
- [ ] Move containerization-specific configuration to `<job>.properties.bosh_containerization`
- [ ] `cf-deployment` + an ops file == role manifest

bkrannich:
> Rename roles to instance_groups
Hooray! Locking into Fissile again recently, I was about to suggest this, too.

> Change secret generator so it supports consuming information from a variables section
For me, the generation and rotation of secrets is a separate concern that I'd suggest trying to pull out from Fissile altogether. The topic of credential rotation needs to be solved in a uniform way anyhow, even for non-containerized workloads.

> cf-deployment + an ops file == role manifest
Another hooray here. I was about to comment "why is the role manifest needed at all?". So maybe the answer to "what is in the ops file?" would address this question, too.

viovanov:
@bkrannich the generation of secrets is not controlled by fissile, but by [this process](https://github.com/SUSE/scf-secret-generator).

It's included as a BOSH Job in [this release](https://github.com/SUSE/scf-helper-release/tree/master/jobs/generate-secrets) and runs as an errand.

The implementation is kube specific, but we can generalize that.

bkrannich:
@viovanov - but why is this a concern that is related to the conversion of a BOSH release into something that can be deployed on K8s? Ultimately, wouldn't you expect to be able to use the K8s primitives for credentials, so secrets and potentially config maps and shouldn't a rotation of those be independent from the pure conversion?

viovanov:
@bkrannich if I understand what you're saying - yes, that's correct. There's no dependency to the secret generator itself.
Secret generation is an "extra" feature, that you don't have to use when deploying.

But we currently use a secret generation mechanism and we want to transition it smoothly, along with everything else.

PR #376, opened 2018-08-02
Rename roles to instance groups

manno:
This implements the first big rename from issue #370 to make the role manifest compatible with BOSH deployment manifests.

* rename `model.Roles` to `InstanceGroups`
* rename `model.RoleJob` to `JobReference`
* rename `roles:` key to `instance_groups:` in role manifest
* rename some related functions, i.e. `IsColocatedContainerRole`
* rename local variables, change comments and error messages
* change tests to tag the instance group with the name "some-group" instead of "role"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 #!/usr/bin/env make
 
 ifeq ($(GIT_ROOT),)
+
+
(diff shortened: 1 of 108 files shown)


viovanov:
This should trigger a major version change

Issue #374, opened 2018-07-22
Create a repo for the NATS "example release"


viovanov:
- [x] Create a github repository for it
- [x] Talk to the guys at SAP to get the latest work
- [ ] Test to see everything works
- [ ] Update the fissile README to point to the new repo

HeavyWombat:
@viovanov We have a Git repo internally which we call something like `fissile-dev` area. It is a Git repo that contains a README and the minimum required configuration to have a simple flow from creating a BOSH relate to generating Docker images and Helm Charts to be deployed on Kubernetes. We use this for doing fissile PRs. We use `nginx` as the BOSH release. Is that comparable to what you have in mind here or do I misunderstand the issue here? Thanks for your feedback.

voelzmo:
@HeavyWombat do you mind sharing the instructions on how to come from bosh release to imager+helm chart? I'd be interested in how much the information differs from what's in https://github.com/SUSE/fissile/blob/master/docs/configuration.md

Should all of that information probably even go into that document or an additional document in this repo?

manno:
@HeavyWombat I think so. We might put these into a sub folder first, like `examples`.  @viovanov  mentioned it would be great to have a release which is clusterable, like nats.

It would be great to test fissile images with minikube.

HeavyWombat:
Ok, let's do it like this: I will create a repo under my user to show what we have plus a lot of disclaimers since we aimed for the smallest possible setup and took every short cut and trick we could find. The [fissile configuration](https://github.com/SUSE/fissile/blob/master/docs/configuration.md) that was mentioned is the full truth, where what we did is more the TL;DR version.

We have a simple one BOSH release version and one with multiple releases that was used for the development of the colocated container change in fissile. If you want you could look into providing a `nats`-based version of the set-up that is clusterable as another branch or so.

HeavyWombat:
@manno @voelzmo This is the repo which @qu1queee and I used for some fissile development work in the past: https://github.com/HeavyWombat/fissile-dev

@smoser-ibm FYI

manno:
@HeavyWombat  this is great. I just created https://github.com/SUSE/nats-containerized-release 
It's still work in progress, but I managed to deploy it to minikube.

PR #380, opened 2018-08-06
Add example releases to README

manno:
This belongs to issue #374 

Maybe we should add SCF, too?
And https://github.com/SUSE/scf-helper-release and https://github.com/SUSE/uaa-fissile-release?


viovanov:
@manno yeah, now that I think about it.
For the NATS one, we should say a bit more, mentioning that it's the preferred release for developing on fissile.

manno:
@viovanov I added more examples and explained the NATS release. Can we merge this?

Issue #383, opened 2018-08-15
fissile build error when commit_hash is a scientific format string


qibobo:
When dev.yml looks like below. The commit_hash is a scientific format string, fissile build will fail:

```
name: scf
version: 0+dev.1
commit_hash: 9018e89
uncommitted_changes: true
jobs:
```
Error:
```
Please allow a long time for mariadb to compile
Error loading dev release information: Error trying to load release scf metadata from YAML manifest /home/vagrant/scf/src/scf-release/dev_releases/scf/scf-0+dev.1.yml: interface conversion: interface {} is float64, not string
Makefile:258: recipe for target 'compile' failed
make: *** [compile] Error 1
```

jandubois:
Issue has been fixed by #384

PR #384, opened 2018-08-15
bug fix: fissile build error when commit_hash is a scientific format string

qibobo:
This PR is to fix issue : https://github.com/SUSE/fissile/issues/383

### DIFF
--- a/model/release.go
+++ b/model/release.go
@@ -27,13 +27,13 @@ type Release struct {
 }
 
 type manifest struct {
-	Name               string            `yaml:"name"`
-	Version            string            `yaml:"version"`
-	CommitHash         string            `yaml:"commit_hash"`
-	UncommittedChanges bool              `yaml:"uncommitted_changes"`
-	Jobs               []interface{}     `yaml:"jobs"`
-	Packages           []interface{}     `yaml:"packages"`
-	License            map[string]string `yaml:"license"`
+	Name               string                        `yaml:"name"`
+	Version            string                        `yaml:"version"`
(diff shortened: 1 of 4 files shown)


model/release.go
```diff
@@ -142,15 +151,14 @@ func (r *Release) loadPackages() (err error) {
...
-	if packages, ok := r.manifest["packages"].([]interface{}); ok {
-		for _, pkg := range packages {
-			p, err := newPackage(r, pkg.(map[interface{}]interface{}))
-			if err != nil {
-				return err
-			}
-
-			r.Packages = append(r.Packages, p)
+	packages := r.manifest.Packages
+	for _, pkg := range packages {
```
mook-as: This could probably be just `range r.manifest.Packages`.

model/release.go
```diff
@@ -123,7 +132,7 @@ func (r *Release) loadJobs() (err error) {
 		}
 	}()
 
-	jobs := r.manifest["jobs"].([]interface{})
+	jobs := r.manifest.Jobs
 	for _, job := range jobs {
```
mook-as: This can just be `for _, job := range r.manifest.Jobs`

Also, if we define `Jobs` as `[]map[interface{}]interface{}`, we won't need the cast on the next line.

jandubois:
> We were thinking of updating the yaml library for this, but it turns out this works fine too

Isn't this better than the updated yaml library? Because this should also work for purely numeric hashes, which would still get the wrong type even with the updated yaml library.

qibobo:
@mook-as updated the PR.

PR #385, opened 2018-08-26
Perform all necessary validation in fissile

viovanov:
Warnings for properties that have different defaults across jobs are removed.
They are not useful when moving to "per job" properties.

### DIFF
--- a/app/validation.go
+++ b/app/validation.go
@@ -295,7 +295,7 @@ func (f *Fissile) checkEnvFileVariables(roleManifest *model.RoleManifest, defaul
 		return false
 	}
 
-	for key, _ := range defaults {
+	for key := range defaults {
 		if !exists(key) {
 			allErrs = append(allErrs, validation.NotFound(key,
 				"Variable from env file not defined in the manifest."))
--- a/app/validation.go
+++ b/app/validation.go
@@ -286,7 +286,7 @@ func (f *Fissile) checkEnvFileVariables(roleManifest *model.RoleManifest, defaul
 	allErrs := validation.ErrorList{}
 
 	exists := func(key string) bool {
-		for _, variable := range roleManifest.Configuration.Variables {
+		for _, variable := range roleManifest.Variables {
 			if variable.Name == key {
 				return true
 			}
@@ -301,34 +301,6 @@ func (f *Fissile) checkEnvFileVariables(roleManifest *model.RoleManifest, defaul
 				"Variable from env file not defined in the manifest."))
 		}
 	}
-
-	return allErrs
-}
-
-// All role manifest templates must use only declared params
-func (f *Fissile) checkAllMustacheTemplatesUseDeclaredVariables() validation.ErrorList {
-	allErrs := validation.ErrorList{}
-
-	return allErrs
-}
-
-// All role manifest templates must be sorted
-func (f *Fissile) checkAllMustacheTemplatesAreSorted() validation.ErrorList {
-	allErrs := validation.ErrorList{}
-
-	return allErrs
-}
-
-// The role manifest must not contain any constants in the global section
-func (f *Fissile) checkNoConstantsInMustacheTemplates() validation.ErrorList {
-	allErrs := validation.ErrorList{}
-
-	return allErrs
-}
-
-// All of the scripts must be used
-func (f *Fissile) checkAllScriptsAreUsedInManifest() validation.ErrorList {
-	allErrs := validation.ErrorList{}
 
 	return allErrs
 }
--- a/app/validation.go
+++ b/app/validation.go
@@ -6,15 +6,13 @@ import (
 
 	"github.com/SUSE/fissile/model"
 	"github.com/SUSE/fissile/validation"
-
-	"github.com/fatih/color"
 )
 
 // validateManifestAndOpinions applies a series of checks to the role
 // manifest and opinions, testing for consistency against each other
 // and the loaded bosh releases. The result is a (possibly empty)
 // array of any issues found.
-func (f *Fissile) validateManifestAndOpinions(roleManifest *model.RoleManifest, opinions *model.Opinions) validation.ErrorList {
+func (f *Fissile) validateManifestAndOpinions(roleManifest *model.RoleManifest, opinions *model.Opinions, defaultsFromEnvFiles map[string]string) validation.ErrorList {
 	allErrs := validation.ErrorList{}
 
 	boshPropertyDefaultsAndJobs := f.collectPropertyDefaults()
@@ -45,9 +43,8 @@ func (f *Fissile) validateManifestAndOpinions(roleManifest *model.RoleManifest, 
 	// opinions
 	allErrs = append(allErrs, checkForDuplicatesBetweenManifestAndLight(lightOpinions, roleManifest)...)
 
-	// All bosh properties in a release should have the same
-	// default across jobs -- WARNING only, not error
-	f.checkBOSHDefaults(boshPropertyDefaultsAndJobs)
+	// All vars in env files must exist in the role manifest
+	allErrs = append(allErrs, f.checkEnvFileVariables(roleManifest, defaultsFromEnvFiles)...)
 
 	// All light opinions should differ from their defaults in the
 	// BOSH releases
@@ -130,13 +127,19 @@ func collectManifestProperties(roleManifest *model.RoleManifest) map[string]stri
 
 	// Per-instance-group properties
 	for _, instanceGroup := range roleManifest.InstanceGroups {
-		for property, template := range instanceGroup.Configuration.Templates {
-			properties[property] = template
+		for _, propertyDef := range instanceGroup.Configuration.Templates {
+			property := propertyDef.Key.(string)
+			template := propertyDef.Value
+
+			properties[property] = fmt.Sprintf("%v", template)
 		}
 	}
 
 	// And the global properties
-	for property, template := range roleManifest.Configuration.Templates {
+	for _, propertyDef := range roleManifest.Configuration.Templates {
+		property := propertyDef.Key.(string)
+		template := propertyDef.Value.(string)
+
 		properties[property] = template
 	}
 
@@ -184,7 +187,10 @@ func checkForDuplicatesBetweenManifestAndLight(light map[string]string, roleMani
 	check := make(map[string]struct{})
 
 	// The global properties, ...
-	for property, template := range roleManifest.Configuration.Templates {
+	for _, propertyDef := range roleManifest.Configuration.Templates {
+		property := propertyDef.Key.(string)
+		template := propertyDef.Value.(string)
+
 		allErrs = append(allErrs, checkForDuplicateProperty("configuration.templates", property, template, light, true)...)
 		check[property] = struct{}{}
 	}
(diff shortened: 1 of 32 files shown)


viovanov:
[(5) fissile: move validation code from scf to fissile](https://trello.com/c/TD9VSPOl/769-5-fissile-move-validation-code-from-scf-to-fissile)

model/roles.go
```diff
@@ -979,6 +994,98 @@ func (j *JobReference) WriteConfigs(instanceGroup *InstanceGroup, lightOpinionsP
...
+	allErrs := validation.ErrorList{}
+
+	previousKey := ""
+
+	for _, templateDef := range roleManifest.Configuration.Templates {
+		key := templateDef.Key.(string)
+
+		if previousKey != "" && previousKey > key {
+			allErrs = append(allErrs, validation.Forbidden(previousKey,
+				fmt.Sprintf("Template key does not sort before '%s'", key)))
```
mook-as: Aren't the two keys reversed here?

model/roles.go
```diff
@@ -979,6 +994,98 @@ func (j *JobReference) WriteConfigs(instanceGroup *InstanceGroup, lightOpinionsP
...
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+
+			if info.IsDir() {
+				return nil
+			}
+
+			if _, ok := usedScripts[path]; !ok {
```
mook-as: This is probably the same as `if !usedScripts[path]`, because all entries are `true`.

model/roles.go
```diff
@@ -1184,14 +1291,22 @@ func validateTemplateUsage(roleManifest *RoleManifest) validation.ErrorList {
...
-	for _, template := range roleManifest.Configuration.Templates {
 		varsInTemplate, err := parseTemplate(template)
 		if err != nil {
 			// Ignore bad template, cannot have sensible
 			// variable references
 			continue
 		}
+
+		if len(varsInTemplate) == 0 {
+			allErrs = append(allErrs, validation.Forbidden(key,
```
mook-as: Should probably still say `configuration.templates.<key>`, so that it's easier to find

app/validation.go
```diff
@@ -130,13 +127,19 @@ func collectManifestProperties(roleManifest *model.RoleManifest) map[string]stri
 
 	// Per-instance-group properties
 	for _, instanceGroup := range roleManifest.InstanceGroups {
-		for property, template := range instanceGroup.Configuration.Templates {
-			properties[property] = template
+		for _, propertyDef := range instanceGroup.Configuration.Templates {
+			property := propertyDef.Key.(string)
+			template := propertyDef.Value
+
+			properties[property] = fmt.Sprintf("%v", template)
```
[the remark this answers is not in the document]
viovanov: It's not always a string. It can be an instance group opinion.

app/validation.go
```diff
@@ -230,50 +239,6 @@ func checkForDuplicateProperty(prefix, property, value string, light map[string]
 	return allErrs
 }
 
-// checkBOSHDefaults reports all properties which were given differing
-// defaults across BOSH releases and the jobs inside.
-func (f *Fissile) checkBOSHDefaults(pd propertyDefaults) {
```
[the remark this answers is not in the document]
viovanov: No, as everything moves to per job properties with almost no namespacing, these will only increase in number.
In your example - what was the other job? Did you find it because of fissile validation?
I don't consider that list to be useful anymore, the number of warnings is too large and we can't do anything about it.

app/validation.go
```diff
@@ -318,3 +280,55 @@ func (f *Fissile) checkLightDefaults(light map[string]string, pd propertyDefault
 
 	return allErrs
 }
+
+// All vars in env files must exist in the role manifest
+func (f *Fissile) checkEnvFileVariables(roleManifest *model.RoleManifest, defaults map[string]string) validation.ErrorList {
+	allErrs := validation.ErrorList{}
+
+	exists := func(key string) bool {
```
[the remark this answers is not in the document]
viovanov: Yeah that would work too and improve performance in an unnoticeable way :).
I think this is easy to read, but I can change it if you insist.

app/validation.go
```diff
@@ -318,3 +280,55 @@ func (f *Fissile) checkLightDefaults(light map[string]string, pd propertyDefault
...
+			allErrs = append(allErrs, validation.NotFound(key,
+				"Variable from env file not defined in the manifest."))
+		}
+	}
+
+	return allErrs
+}
+
+// All role manifest templates must use only declared params
+func (f *Fissile) checkAllMustacheTemplatesUseDeclaredVariables() validation.ErrorList {
```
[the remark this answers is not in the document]
viovanov: Ah yes, this is a duplicate - this check is already implemented.
I'll remove this.

model/roles.go
viovanov: No, an error would read:
```
foo.bar: Template key does not sort before bar.foo.
```

model/roles.go
viovanov: I think that makes it harder to find because I can't copy paste and search.

viovanov:
@mook-as I'd like yaml references too.
I think it's out of scope though. And maybe we should implement it after we have a manifest that's bosh compatible.

model/roles.go
andreas-kupries: Not necessarily if `usedScripts` is a map. Then the entry may not exist, and the code as-is treats this non-presence as `false`. Your proposed code may panic instead ?

app/validation.go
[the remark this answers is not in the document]
mook-as: The other job was in router somewhere, and yes it was found via fissile validation, IIRC. (Sebi was looking at making it customizable)

I'm okay with not caring about it, though! :D

app/validation.go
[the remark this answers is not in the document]
mook-as: Nah, I thought thought that was easier to read; but if you find this better, that's fine too.

model/roles.go
mook-as: Ah, right, okay :+1:

model/roles.go
mook-as: No, the zero value of a boolean is false. https://play.golang.org/p/L1PBhJI9v2o
This would only panic if the map was not allocated.

model/roles.go
mook-as: You can just… copy only the part after the dot? Compare if a property is overridden in one job, _and_ it's in the global section.

But whatever.

model/roles.go
andreas-kupries: Ok. Retracting my comment

app/validation.go
```diff
@@ -130,13 +127,19 @@ func collectManifestProperties(roleManifest *model.RoleManifest) map[string]stri
...
+
+			properties[property] = fmt.Sprintf("%v", template)
 		}
 	}
 
 	// And the global properties
-	for property, template := range roleManifest.Configuration.Templates {
+	for _, propertyDef := range roleManifest.Configuration.Templates {
+		property := propertyDef.Key.(string)
+		template := propertyDef.Value.(string)
```
mook-as: This is always castable right?

app/validation_test.go
```diff
@@ -133,3 +133,138 @@ func TestValidationHash(t *testing.T) {
 	}
 	assert.Len(t, errs, len(allExpected))
 }
+
+func TestMandatoryDescriptions(t *testing.T) {
```
mook-as: Is this `t.Parallel()`able?

Same with the other new tests. (The old ones too, really, but we don't care about those right now)

model/roles.go
```diff
@@ -354,6 +360,102 @@ func (m *RoleManifest) SelectInstanceGroups(roleNames []string) (InstanceGroups,
...
+}
+
+// validateSortedTemplates tests that all templates are sorted in alphabetical order
+func validateSortedTemplates(roleManifest *RoleManifest) validation.ErrorList {
+	allErrs := validation.ErrorList{}
+
+	previousKey := ""
+
+	for _, templateDef := range roleManifest.Configuration.Templates {
+		key := templateDef.Key.(string)
```
mook-as: I just realized… we don't have anything to validate that `templateDef.Key` is castable to a string (and that `.Value` is also a string), or that the instance-group variants have string keys (and possibly non-string values, so no check there).

test-assets/role-manifests/app/tor-invalid-environ-script-reference.yml
```diff
@@ -0,0 +1,42 @@
+# This role manifest should pass validation with no issues
```
mook-as: This comment seems suspicious. (Also the same for the other similar new manifests)

test-assets/role-manifests/app/tor-validation-ok.yml
```diff
@@ -2,15 +2,8 @@
...
 - name: myrole
-  environment_scripts:
-  - environ.sh
-  - /environ/script/with/absolute/path.sh
   scripts:
   - myrole.sh
-  - /script/with/absolute/path.sh
-  post_config_scripts:
-  - post_config_script.sh
-  - /var/vcap/jobs/myrole/pre-start
```
mook-as: Probably leave the absolute paths in? Unless you have the equivalent checks elsewhere…

app/validation.go
viovanov: Globals yes.
They are not allowed to be constants - they need to contain a string.

app/validation_test.go
viovanov: Should be, yeah

model/roles.go
viovanov: does yaml allow non-string keys?
.Value is a string because we don't allow constants in the global section.
Not sure if I understand correctly.

test-assets/role-manifests/app/tor-invalid-environ-script-reference.yml
viovanov: Yep

test-assets/role-manifests/app/tor-validation-ok.yml
viovanov: Yes, these are tested for explicitly elsewhere.
They were copy pasted I assume in here, and they don't add to the tests where this is used.

model/roles.go
mook-as:
```diff
diff --git a/test-assets/role-manifests/app/tor-missing-description.yml b/test-assets/role-manifests/app/tor-missing-description.yml
index 0caa290..4d3d971 100644
--- a/test-assets/role-manifests/app/tor-missing-description.yml
+++ b/test-assets/role-manifests/app/tor-missing-description.yml
@@ -23,6 +23,7 @@ configuration:
     properties.tor.hashed_control_password: '((={{ }}=)){{PELERINUL}}'
     properties.tor.hostname: '((FOO))'
     properties.tor.private_key: '((#BAR))((HOME))((/BAR))'
+    3: 5
 variables:
 - name: BAR
   options:
```
```
==> Testing 
?       github.com/SUSE/fissile [no test files]
--- FAIL: TestMandatoryDescriptions (0.06s)
panic: interface conversion: interface {} is int, not string [recovered]
        panic: interface conversion: interface {} is int, not string

goroutine 69 [running]:
testing.tRunner.func1(0xc4201a60f0)
        /home/marky/opt/golang/current/src/testing/testing.go:711 +0x5d9
panic(0xaf3620, 0xc42037cbc0)
        /home/marky/opt/golang/current/src/runtime/panic.go:491 +0x2a2
github.com/SUSE/fissile/model.(*InstanceGroup).calculateRoleConfigurationTemplates(0xc4202682a0)
        /home/marky/src/work/src/github.com/SUSE/fissile/model/instance_groups.go:485 +0x903
```

And I'm saying we don't have any sanity checking that `.Value` _is_ a string; I can do `foo: [ a, b, c ]` and cause a panic.

viovanov: Yes, I agree.
This has been always the case afaict. 
https://www.pivotaltracker.com/story/show/160305258

mook-as: No, previously the struct was a `map[string]string`, and there was no casting (so we can't cast it wrong).

PR #386, opened 2018-08-29
kube: services: use the port number for targetPort definitions

mook-as:
When multiple port definitions exist on a pod with the same containerPort (but different external ports), we need to end up using the first one for the service definition as kubernetes has trouble correctly modifying the port definitions on upgrades.  The easiest way to make that work is to just use the numeric port number for the service targetPort, which will end up referring to whichever container port definition that works.


mook-as:
The upgrade was only busted because the port names weren't updated. By using numbers, kube will still be able to find the ports (by number, instead of by name), so it should work fine :)

PR #391, opened 2018-08-31
Support for referencing releases in manifests

viovanov:
Depends on #385 

This allows things like this:
```
---
releases:
- name: "capi"
  version: "1.66.0"
  url: "https://bosh.io/d/github.com/cloudfoundry/capi-release?v=1.66.0"
  sha1: "4236495d4a047fb9396996aba3daea9e0cb3f1bf"
...
instance_groups:
- name: syslog-scheduler
  environment_scripts:
...
```

When fissile encounters something like this, it will download the release, unpack it and use it as a final release.


model/release.go
```diff
@@ -11,6 +11,14 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+// ReleaseRef represents a reference to a BOSH release from a manifest
+type ReleaseRef struct {
+	Name    string `yaml:"name"`
+	URL     string `yaml:"url"`
+	Sha1    string `yaml:"sha1"`
```
jandubois: `Sha1` should be `SHA1` according to Go naming guidelines.

PR #395, opened 2018-09-13
In per-job templates the values may be boolean or numbers too

jandubois:


jandubois:
Made it `WIP` to maybe add tests, but I need a PR right now to get a checks build from CI

viovanov:
This also avoids panics when global templates are not strings, and returns an appropriate error message.

PR #396, opened 2018-09-13
Remove support for instance groups of the `docker` type

manno:
We don't support docker roles anymore. However they create problems when moving the run property from the `InstanceGroup` to the `JobReferences` properties. `Run` is required on at least one job and docker roles don't have jobs (?). So lets remove them.

Unsure if these are all the places.


jandubois:
I don't want to merge this until after #391 has been merged. Unfortunately I did not finish reviewing it today.

manno:
I rebased onto #391 and updated this PR with 1cdcc7d

PR #397, opened 2018-09-14
Move containerization properties below BOSH properties

manno:
* iterate jobs for run and colocated containers properties
* merge `Run` properties from jobs
* extract methods to structs
* remove `Run.Environment`, only used by docker instance groups
* remove `ObjectAnnotations`

https://www.pivotaltracker.com/story/show/160013643


viovanov:
@manno we'll need a PR for changing SCF/UAA before we can merge this

PR #401, opened 2018-09-18
Refactoring before starting on "replace templates with job properties"

manno:


manno:
* After splitting the models across more files, tests were not moved to corresponding files
* Adding more functionality to the `JobReference` makes that model more prominent. The relationship to `Job` is not clear. `Job` is also embedded into `JobReference`. Renaming this to make clear `Job` data comes from the job spec of a release, while `JobReference`  is the data from the deployment manifest. Calling it a reference makes sense in the regard, that it references a job spec from a release by ReleaseName.
* The `Job.Properties` are actually the property description from the releases spec. Renaming it because JobReference gets its own `Properties` now.

PR #414, opened 2018-09-29
Fix link target name for symbolic links in tar archive

### LINKED ISSUE #413: New compilation cache code creates broken archives
@VIovanov @Tibeos:

The problem is that relative symlinks are turned into absolute symlinks:

```
# ls -l /var/vcap/packages/postgres-9.6.6/lib/libpq.so.5
lrwxrwxrwx 1 root root 140 Sep 28 19:42 /var/vcap/packages/postgres-9.6.6/lib/libpq.so.5 -> /home/vagrant/.fissile/compilation/4c162e7204bd9cd84142a77c6de81eaf154bf16a/64abaa2d40d8a9482fa1a4ee8b9e8301b6898dc8/compiled/lib/libpq.so.5
```

This should have been a simple relative symlink:

```
# ls -l /var/vcap/packages/postgres-9.6.6/lib/libpq.so.5
lrwxrwxrwx 1 root root 12 Sep 11 01:35 /var/vcap/packages/postgres-9.6.6/lib/libpq.so.5 -> libpq.so.5.9
```

The problem arises because the absolute symlink points into the fissile compilation cache directory (which doesn't exist inside the container) and not to the actual install location.

This seems to be a known limitation of the archiver package: https://github.com/mholt/archiver/issues/74

jandubois:
Attempt to fix #413

### DIFF
--- a/vendor/github.com/mholt/archiver/tar.go
+++ b/vendor/github.com/mholt/archiver/tar.go
@@ -140,7 +140,14 @@ func tarFile(tarWriter *tar.Writer, source, dest string) error {
 			return fmt.Errorf("error walking to %s: %v", path, err)
 		}
 
-		header, err := tar.FileInfoHeader(info, path)
+		link := path
+		if (info.Mode() & os.ModeSymlink) != 0 {
+			link, err = os.Readlink(path)
+			if err != nil {
+				return fmt.Errorf("%s: readlink: %v", path, err)
+			}
+		}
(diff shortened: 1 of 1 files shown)


jandubois:
I have submitted a more comprehensive upstream PR for this issue: https://github.com/mholt/archiver/pull/92.

If we merge this temporary fix, then we should open a new issue to update dependencies once upstream has been updated.

PR #417, opened 2018-10-02
Prefix the global entities clusterrole(binding) with namespace information.

andreas-kupries:
Keep the definitions in different charts apart from each other.
Updated tests.

Ref: https://trello.com/c/KzfK2xOh/822-13-rc-deploy-issue-with-psp-cluster-roles

:warning: While it passes the unit tests this is still to be tested with scf.

### DIFF
--- a/kube/rbac.go
+++ b/kube/rbac.go
@@ -52,7 +52,8 @@ func NewRBACAccount(name string, account model.AuthAccount, settings ExportSetti
 	}
 
 	binding := newTypeMeta("rbac.authorization.k8s.io/v1", "ClusterRoleBinding", blockPSP)
-	binding.Add("metadata", helm.NewMapping("name", fmt.Sprintf("%s-binding-psp", name)))
+	binding.Add("metadata", helm.NewMapping("name",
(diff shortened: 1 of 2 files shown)


kube/rbac.go
```diff
@@ -52,15 +52,16 @@ func NewRBACAccount(name string, account model.AuthAccount, settings ExportSetti
 	}
 
 	binding := newTypeMeta("rbac.authorization.k8s.io/v1", "ClusterRoleBinding", blockPSP)
-	binding.Add("metadata", helm.NewMapping("name", fmt.Sprintf("%s-binding-psp", name)))
+	binding.Add("metadata", helm.NewMapping("name",
+		authCRBindingName(name, settings.CreateHelmChart)))
```
jandubois: We generally try to pass just `settings` into the lower level functions, and let them check for `settings.CreateHelmChart` themselves.

andreas-kupries: Addressed in commit 8f8210a40edcde2491a1d2e44f13d68b1fefaa27

PR #418, opened 2018-10-03
Validate that scripts exist and that all scripts are used; also, that the path starts with scripts/

mook-as:
Related changes:
- model.LoadRoleManifest() now takes a struct for options (so that we can skip optional things)
- model.LoadRoleManifest() can have an option to ignore missing scripts (so that not all tests have to have scripts in-tree to pass)
- moved a bunch of validation from `model/roles.go` to `model/validation.go` (no functional change)

This is the only checks from scf.git that isn't currently in fissile, and that we can move over.  The remaining checks are all about specific scripts, or not valid anymore:
- all roles using HA configuration have `configure-HA-hosts.sh` in an environment script
- all BOSH roles (… everything) includes `forward_logfiles.sh`
- no non-docker roles (… everything, sinec docker roles no longer exist) have a `run.env` key
- multiple releases do not have the same property with different values (this turns out to always be violated in practice, and thus basically had no value)


model/validation.go
```diff
@@ -0,0 +1,907 @@
...
+
+	for _, instanceGroup := range roleManifest.InstanceGroups {
+		for scriptType, scriptList := range map[string][]string{
+			"script":             instanceGroup.Scripts,
+			"environment script": instanceGroup.EnvironScripts,
+			"post config script": instanceGroup.PostConfigScripts,
+		} {
+			for _, script := range scriptList {
+				if filepath.IsAbs(script) {
+					continue
```
andreas-kupries: Do we actually allow absolute script paths ?
Should we maybe print a warning about the trouble this can be, even if not an error ?

IMHO this is special enough to require an informational comment.

mook-as: We do! Because (at least at some point) we manually added scripts from `/var/vcap/packages`, which fissile has no knowledge about.

andreas-kupries: Ok. Can we document that, please?

mook-as: Comment added in 97c7c9d; let me know if you have a different kind of document in mind.

andreas-kupries: Your comment looks ok.

PR #420, opened 2018-10-03
Replace skiff-role-name with app.kubernetes.io/component

jandubois:
We'll keep the skiff-role-name label around for a while in case it is being used by end users.


jandubois:
Do NOT merge yet; we need corresponding changes in configgin and SCF first.

PR #426, opened 2018-10-10
helm+kube: default to not allowing privilege escalation

mook-as:
When using pod security policies, it is likely that (for the unprivileged policy at least) we are not allowed to escalate privileges. This means that pods will fail to start if they ask for escalation.

For some crazy reason, kube seems to default `allowPrivilegeEscalation` to true when not supplied (when fetching the pod metadata, not when running); this means that attempting to patch the pod (to update the configgin annotations) will fail because even a no-op patch is considered an escalation.  Fix this by explicitly marking the pods as not requiring the escalation.


jandubois:
I rebased this PR on top of latest `develop` to make it build with latest `goimports` again.

PR #436, opened 2018-10-26
Add test for util.DownloadFile

mudler:
Hi,

While I was digging the codebase, noticed that ```DownloadFile``` missed a test file. The test is actually creating a small http server so it can be performed in network sandboxed environments.


viovanov:
Thanks @mudler

PR #438, opened 2018-11-01
Add nodeAffinity template for roles that don't specify podAntiAffinity

jandubois:
[trello#m3yGuxn2]


kube/deployment.go
```diff
@@ -62,23 +62,23 @@ func addAffinityRules(instanceGroup *model.InstanceGroup, spec *helm.Mapping, se
...
 
-		annotations.Sort()
+	meta := spec.Get("template", "metadata").(*helm.Mapping)
+	if meta.Get("annotations") == nil {
+		meta.Add("annotations", helm.NewMapping())
+		meta.Sort()
 	}
+	annotations := meta.Get("annotations").(*helm.Mapping)
+
+	annotations.Sort()
```
mook-as: So this chunk ensures that there's a `.template.metadata.annotations`, which might be empty? Does this actually do anything _useful_?

I recognize this is old code but still…

andreas-kupries: Looking over the code I believe that we could indeed add a `return` after line 77 (`meta.Sort()`) without changing results.

Lines 79-81 are, I believe, needed only if we had annotations when coming into this function. These have to be sorted. When it is us here adding an empty set of annotations, sorting is irrelevant.

andreas-kupries: I confirm that the proposed change passes the fissile unit tests.

andreas-kupries: Looks ok in the SCF chart too. No changes compared to without the change. And the PR as-is inserts the proper affinity templating (affected roles: autoscaler-*, blobstore, credhub).

PR #439, opened 2018-11-02
Adjust /var/vcap layout for bpm compatibility

jandubois:
The symlinks from /var/vcap/packages to ../packages-src do not resolve inside a bpm container, so move the packages over instead.

bosh also creates a number of convenience symlinks under /var/vcap/data that this commit simply emulates.


jandubois:
This change is necessary for me to get the nats-container-deployment working because the latest nats-release master unconditionally uses bpm.

PR #444, opened 2018-11-08
 kube/pod.go: force allowPrivilegeEscalation if privileged

mook-as:
Otherwise kube will reject the pod with:

  create Pod … in StatefulSet … failed error: Pod … is invalid:
   … cannot set `allowPrivilegeEscalation` to false and
   `privileged` to true

### DIFF
--- a/kube/pod.go
+++ b/kube/pod.go
@@ -516,7 +516,6 @@ func getSecurityContext(role *model.InstanceGroup, createHelmChart bool) helm.No
 
 	sc := helm.NewMapping()
 	allowPrivileged := role.PodSecurityPolicy() == model.PodSecurityPolicyPrivileged
-	sc.Add("allowPrivilegeEscalation", allowPrivileged)
 
 	capabilities := role.Run.Capabilities
 	if createHelmChart {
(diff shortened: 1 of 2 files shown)


kube/pod.go
```diff
@@ -529,10 +528,17 @@ func getSecurityContext(role *model.InstanceGroup, createHelmChart bool) helm.No
 		// Complete the context, with conditional privileged mode
 		sc.Add("privileged", helm.NewNode(true, helm.Block(hasAll)))
 		sc.Add("capabilities", cla)
+		if allowPrivileged {
+			sc.Add("allowPrivilegeEscalation", allowPrivileged)
+		} else {
+			// We ned to allow privilege escalation if if want privileged mode
+			sc.Add("allowPrivilegeEscalation", fmt.Sprintf("{{ %s }} true {{ else }} false {{ end }}", hasAll))
```
jandubois:
```suggestion
			sc.Add("allowPrivilegeEscalation", fmt.Sprintf("{{ %s -}} true {{- else -}} false {{- end }}", hasAll))
```

Don't feel strongly, but my OCD makes me suggest this change... 😄

PR #446, opened 2018-11-08
 Include config vars without default value in kube config

jandubois:


andreas-kupries (approved):
It does look ok to me.

@mook-as This is about https://trello.com/c/4yltS6ZB/817-3-remove-env-files-from-scf-uaa where some variables were lost from kube config when the env files did not supply a default fro them.

mook-as (commented):
Looks inoffensive. Still trying to make sure this is the _correct_ change, though…

Why do we now want an empty env var (as opposed to unset)?

jandubois:
> Why do we now want an empty env var (as opposed to unset)?

So we can modify the value in `kube_overrides.rb` in SCF

PR #448, opened 2018-11-16
run.sh: Do the move dance only when BPM is present; chown, too

mook-as:
There are odd permissions issues on Jenkins sometimes with the packages directories.  They seem to be resolved by changing the owner of the files so that `vcap` can read them.

This is a workaround that should be removed once we drop the packages layer.


mook-as (commented):
Attempting to ensure the PR cannot be merged (by requesting a change).

mook-as:
Okay, this appears to resolve the issues I'm seeing on Jenkins; not particularly _happy_ about it, though. (This is mostly working for me, I think, because none of the things I use have BPM yet.)

Please check that this does what you expect in the BPM world.

jandubois (approved):

PR #449, opened 2018-11-19
Add support for compiling on a remote docker

viovanov:


Gopkg.toml
```diff
@@ -71,7 +75,7 @@
 
 [[constraint]]
   name = "github.com/mholt/archiver"
-  version = "2.1.0"
+  version = "3.1.0"
```
jandubois: This change will make us drop a local patch to the vendor tree: https://github.com/cloudfoundry-incubator/fissile/pull/414/files

We have an open issue to update this dependency once the upstream PR has been merged: https://github.com/cloudfoundry-incubator/fissile/issues/416

I will need to rebase my PR on top of a restructuring of the code from v2 to v3, and then still need to add additional code for Windows support before we can expect this to be merged.

So for now we should not updated our vendored version of mholt/archiver.

docker/docker.go
```diff
@@ -529,12 +536,13 @@ func (d *ImageManager) RunInContainer(opts RunInContainerOpts) (exitCode int, co
...
 	for src, dest := range opts.Mounts {
-		if _, ok := opts.Volumes[src]; ok {
+		var ok bool
+		if _, ok = opts.Volumes[src]; ok {
 			// Attempt to mount a volume; use the generated name
 			src = fmt.Sprintf("volume_%s_%s", opts.ContainerName, src)
 		}
 		mountString := fmt.Sprintf("%s:%s", src, dest)
-		if dest == ContainerInPath {
+		if dest == ContainerInPath && !ok {
```
jandubois, line 553: Why is the `&& !ok` needed? Maybe explain in a comment?

docker/docker.go
```diff
@@ -566,6 +574,75 @@ func (d *ImageManager) RunInContainer(opts RunInContainerOpts) (exitCode int, co
 	}
 	attached <- <-attached
 
+	// Stream files within the container just before starting
+	fsWithSymlinks := fs.NewFileSystem()
+	fsWithSymlinks.KeepSymlinks(true)
+	for src, dst := range opts.StreamIn {
```
jandubois: `dst` should be `dest`. See also usage just 40 lines above.

docker/docker.go
```diff
@@ -566,6 +574,75 @@ func (d *ImageManager) RunInContainer(opts RunInContainerOpts) (exitCode int, co
...
+		tarstream := tarstream.New(fsWithSymlinks)
```
jandubois: Maybe use `tarStream` to avoid using the same name for a package and a variable?

docker/docker.go
```diff
@@ -566,6 +574,75 @@ func (d *ImageManager) RunInContainer(opts RunInContainerOpts) (exitCode int, co
...
+			Path:        dst,
+		})
+		if err != nil {
+			return -1, container, fmt.Errorf("Error running in container: %s. Error streaming data into container: %s", container.ID, err)
+		}
+	}
+
+	// Function for streaming files out of the container
+	streamOutFiles := func() error {
+		for src, dst := range opts.StreamOut {
```
jandubois: `dest`, see above.

docker/docker.go
```diff
@@ -566,6 +574,75 @@ func (d *ImageManager) RunInContainer(opts RunInContainerOpts) (exitCode int, co
...
 
+	// Stream files within the container just before starting
+	fsWithSymlinks := fs.NewFileSystem()
+	fsWithSymlinks.KeepSymlinks(true)
+	for src, dst := range opts.StreamIn {
+		tarstream := tarstream.New(fsWithSymlinks)
+		r, w := io.Pipe()
+
+		go func() {
+			tarErr := tarstream.CreateTarStream(src, false, w)
```
jandubois: Why is this not just `err`?

docker/docker.go
```diff
@@ -566,6 +574,75 @@ func (d *ImageManager) RunInContainer(opts RunInContainerOpts) (exitCode int, co
...
+	}
+
+	// Function for streaming files out of the container
+	streamOutFiles := func() error {
+		for src, dst := range opts.StreamOut {
+			tarstream := tarstream.New(fsWithSymlinks)
+			r, w := io.Pipe()
+
+			go func() {
+				downloadErr := d.client.DownloadFromContainer(container.ID, dockerclient.DownloadFromContainerOptions{
```
jandubois: `err`?

docker/docker.go
```diff
@@ -566,6 +574,75 @@ func (d *ImageManager) RunInContainer(opts RunInContainerOpts) (exitCode int, co
...
+			err := tarstream.ExtractTarStream(dst, r)
+			if err != nil {
+				return err
+			}
+
+			// Docker will include the directory in the output tar stream so
+			// we need to move things arround
+			sourceDirectoryName := filepath.Base(src)
+			streamedOutputDir := filepath.Join(dst, sourceDirectoryName)
+			tmpDstDir := filepath.Join(filepath.Dir(dst), fmt.Sprintf("%s-tmp", filepath.Base(dst)))
```
jandubois: `tempDestDir`?

docker/docker.go
```diff
@@ -605,11 +689,18 @@ func (d *ImageManager) RunInContainer(opts RunInContainerOpts) (exitCode int, co
 	} else {
 		exitCode = -1
 	}
+
+	// Stream files out of the container
+	streamErr := streamOutFiles()
```
jandubois: `err`?

docker/docker.go
```diff
@@ -605,11 +689,18 @@ func (d *ImageManager) RunInContainer(opts RunInContainerOpts) (exitCode int, co
 	} else {
 		exitCode = -1
 	}
+
```
jandubois, line 707: Why are you still streaming the files if `err` isn't `nil`?

docker/docker.go
```diff
@@ -605,11 +689,18 @@ func (d *ImageManager) RunInContainer(opts RunInContainerOpts) (exitCode int, co
...
+	// Stream files out of the container
+	streamErr := streamOutFiles()
+	if streamErr != nil {
+		return exitCode, container, fmt.Errorf("Error running in container: %s. Error streaming data out of container: %s", container.ID, streamErr)
+	}
 	closeFiles()
+
```
jandubois, line 709: This function has named return values: `(exitCode int, container *dockerclient.Container, err error)`. Can we get rid of the names; they don't seem to be used?

jandubois (changes requested):
I haven't actually tested the code streaming to a remote docker instance.

Most of the review feedback is stylistic only, except for the need to keep the support for symlinks in tar files in the vendored `mholt/archiver` package.

Gopkg.toml
viovanov: ah.. I have to revert a bunch of code
the openshift package I used for streaming has symlink support, maybe we should consider switching if your PR is stuck

docker/docker.go
viovanov: `exitCode` is used

docker/docker.go
```diff
@@ -586,7 +671,15 @@ func (d *ImageManager) RunInContainer(opts RunInContainerOpts) (exitCode int, co
 		closeFiles()
 		if err != nil {
 			exitCode = -1
+			return exitCode, container, err
```
jandubois, line 674:
```suggestion
			return -1, container, err
```

and delete the assignment above, for consistency with other usage in the same function.

jandubois (approved):

PR #456, opened 2018-12-05
support diffing final releases.

### LINKED ISSUE #455: `fissile diff` only works with already unpacked release directories and not with URLs
I expected to be able to run 
```
$ fissile diff --release https://bosh.io/d/github.com/cloudfoundry/nats-release?v=23,https://bosh.io/d/github.com/cloudfoundry/nats-release?v=24
```

But that doesn't work:
```
dev config diff: error loading release information: Error loading dev release information: Path /Users/jan/go/src/code.cloudfoundry.org/fissile/https:/bosh.io/d/github.com/cloudfoundry/nats-release?v=23 (release directory) does not exist
```

mook-as:
- release-resolver: unify loading via releaseRefs
- release-resolver: use temporary files for URLs
- release-resolver: look up release info from final release tarballs
- cmd: skip normalizing release paths that look like URLs

Fixes #455.

I don't like this much yet; there's a bit with a hard `://` check that smells a bit. It's also a lot of change in code I'm not all that familiar with.

:bug:Bug: this seems to accidentally move the final releases dir, which means we'll need to re-download things again.

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -724,14 +724,16 @@ func (f *Fissile) GetDiffConfigurationBases(releasePaths []string, cacheDir stri
 	if len(releasePaths) != 2 {
 		return nil, fmt.Errorf("expected two release paths, got %d", len(releasePaths))
 	}
-	defaultValues := []string{}
-	releases, err := releaseresolver.LoadReleasesFromDisk(model.ReleaseOptions{
-		ReleasePaths:    releasePaths,
-		ReleaseNames:    defaultValues,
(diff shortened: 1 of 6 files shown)


jandubois (changes requested):
I'm going to block this because we are going to change how you specify release refs from the commandline for the `fissile build release-images` command (using `--name`, `--sha1`, `--url`, and `--version` options, so we can check if we have them in the local cache or not). I just discussed this earlier today with Vlad.

This change should then use the same mechanism.

PR #457, opened 2018-12-12
Add support to align with Istio conventions

zhanggbj:
- Add an option for fissile to indicate whether to generate kube yaml and
 helm chart that have Istio conventions.
- Fissile can choose whether to make a role managed by Istio by role tags

[Story 162262084](https://www.pivotaltracker.com/story/show/162262084)


cmd/build-helm.go
```diff
@@ -13,6 +13,7 @@ var (
 	flagBuildHelmUseCPULimits    bool
 	flagBuildHelmTagExtra        string
 	flagBuildHelmAuthType        string
+	flagBuildHelmIstioComplied   bool
```
jandubois: I don't think there is any reason to make this behavior optional.

Unless I misunderstand, the only side-effect of always enabling this functionality is that all roles would get an unused `sidecar.istio.io/inject=false` annotation when Istio isn't being used.

Keeping the code simpler (and avoiding a user-visible option) seems preferable to me.

kube/utils.go
```diff
@@ -11,6 +11,11 @@ import (
 const (
 	// RoleNameLabel is the recommended kube label to specify the rolename
 	RoleNameLabel = "app.kubernetes.io/component"
+	// AppNameLabel is to add contextual information in distributed tracing for Istio
+	AppNameLabel = "app"
+	// AppVersionLable is to indicate the version of app. It is used to add contextual information in
+	// distributed tracing and the metric telemetry collected by Istio
+	AppVersionLable = "version"
```
jandubois: Should be `AppVersionLabel`

kube/utils.go
```diff
@@ -21,26 +26,40 @@ func newTypeMeta(apiVersion, kind string, modifiers ...helm.NodeModifier) *helm.
...
 func newKubeConfig(settings ExportSettings, apiVersion, kind, name string, modifiers ...helm.NodeModifier) *helm.Mapping {
 	labels := helm.NewMapping(RoleNameLabel, name) // "app.kubernetes.io/component"
+	if settings.IstioComplied {
+		labels.Add(AppNameLabel, name)
+	}
+
 	if settings.CreateHelmChart {
 		// XXX skiff-role-name is the legacy RoleNameLabel and will be removed in a future release
 		labels.Add("skiff-role-name", name)
+		// app: XXX is used by Istio
```
jandubois: Not sure what this comment is saying.

jandubois (changes requested):

cmd/build-helm.go
zhanggbj: @jandubois yes, I agree that the only side-effect for now is about the annotation for injection. 

As this is the first PR for Istio integration,  the next PRs(in progress) will be in scf side, and these PRs should work together to finally support Integration. So I just want to keep kube yaml/helm chart "clean" generated by Fissile, although these annotation & labels don't affect any functions : )

And last time we had a discussion with @viovanov @manno , and I remember they mentioned it's better to have a feature option here

How about we keep the option here, and after the scf PRs merged and maybe Istio support is true by default and then we remove this option? Thanks!

kube/utils.go
zhanggbj: Thank you! I'll update in next patch.

kube/utils.go
zhanggbj: Will update in next patch.Thanks!

cmd/build-helm.go
jandubois: > How about we keep the option here, and after the scf PRs merged and maybe Istio support is true by default and then we remove this option? Thanks!

I talked with @mook-as about this, and we both prefer not to have this option. It makes sense to have a feature-flag when the new features changes existing behavior, but that isn't the case here.

If you really want to keep the flag though, then we need to change the name; it is very hard to tell what `--istio-complied` means. At first it looked like a misspelling of `istio compiled`, but eventually I realized it means maybe `istio compliant`. Maybe `--use-istio` would be a better name? And use similar changes to the internal variable names.

It is up to you, but again, I would rather see all of this removed and have Istio support always enabled.

cmd/build-helm.go
```diff
@@ -126,5 +129,12 @@ func init() {
 		"Sets the Kubernetes auth type",
 	)
 
+	buildHelmCmd.PersistentFlags().BoolP(
+		"istio-complied",
+		"",
+		false,
+		"Comply with Istio requirements when generating helm chart",
```
jandubois: If you change the option to `--use-istio` maybe use a description like

```
Add istio labels and annotations to generated helm chart
```

cmd/build-helm.go
zhanggbj: @jandubois 
Thanks a lot for your suggestions :-) After a second thought, I removed the option, so the labels and annotations for Istio will always be there.  Would you please help review again? Thanks!

viovanov:
After quickly reading the conversation - I want to make sure there's an option from `values.yaml` to disable the annotations.
Agreed that making the choice at build time is not useful, but we _must_ be able to make the choice at runtime, as discussed.

zhanggbj:
Hi @viovanov @jandubois,

I just updated the PR and added an option `use-istio` in `values.yml` for users to specify whether to add Istio related labels and annotations in helm chart at runtime. As these labels and annotations won't affect original functions, so I set the default to `true`. Would you please help take a review again? Thanks!

jandubois (approved):

PR #458, opened 2018-12-18
Compile and build per-release docker images

jandubois:
[#161745378]


jandubois:
Marked `WIP` because still missing tests. Only created the PR to get a CI build for testing that the existing functionality still works for SCF builds.

jandubois:
Sample command for building a test image:

```
$FISSILE build release-images \
         --repository test \
         --stemcell splatform/fissile-stemcell-opensuse \
         --work-dir "${HOME}/tmp" \
         --name nats \
         --sha1 65d8efc68d86a03d0f51562e16763075c37e419d \
         --url https://bosh.io/d/github.com/cloudfoundry/nats-release?v=24 \
         --version 24
```

app/fissile_test.go
```diff
@@ -29,18 +29,14 @@ func TestCleanCacheEmpty(t *testing.T) {
...
-	releasePath := filepath.Join(workDir, "../test-assets/ntp-release")
-
 	f := NewFissileApplication(".", ui)
-	err = f.LoadManifest(
-		roleManifestPath,
-		[]string{releasePath},
-		[]string{""},
-		[]string{""},
-		filepath.Join(workDir, "../test-assets/bosh-cache"))
+	f.Options.RoleManifest = filepath.Join(workDir, "../test-assets/role-manifests/app/no-instance-groups/no-instance-groups.yml")
```
f0rmiga, line 33: `filepath.Join` should take each part of the provided path as separate arguments. The way it is constructed currently I'm not sure if it would work on Windows. In this case, it should be:
```
filepath.Join(workDir, "..", "test-assets", "role-manifests", "app", "no-instance-groups", "no-instance-groups.yml")
```

app/build_images.go
```diff
@@ -0,0 +1,225 @@
...
+	NoBuild                  bool
+	OutputDirectory          string
+	PatchPropertiesDirective string
+	Roles                    []string
+	Stemcell                 string
+	StemcellID               string
+	TagExtra                 string
+}
+
+// BuildImages builds all role images using releases
```
f0rmiga: [nit] Missing punctuation. Here and elsewhere.

app/build_images.go
```diff
@@ -0,0 +1,225 @@
...
+func (f *Fissile) BuildImages(opt BuildImagesOptions) error {
+	err := f.LoadManifest()
+	if err != nil {
+		return err
```
f0rmiga, line 34: It is a good practice to propagate the error with a meaningful message. How about:
```go
return fmt.Errorf("failed to build images: %v", err)
```

You can save yourself the burden of doing `fmt.Errof` every time by creating a custom error type. Up to you!

```go
type buildImagesError struct {
    err error
}

func (e *buildImagesError) Error() string {
    return fmt.Sprintf("failed to build images: %v", e.err)
}
```

Then you can use like this:

```go
if err != nil {
    return &buildImagesError{err}
}
```

app/build_images.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	if f.Manifest == nil || len(f.Manifest.LoadedReleases) == 0 {
+		return fmt.Errorf("Releases not loaded")
```
f0rmiga, line 37: [nit] capitalizing errors?

app/build_images.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return err
+	}
+
+	imageName, err := packagesImageBuilder.GetImageName(f.Manifest, instanceGroups, f)
+	if err != nil {
+		return err
+	}
+
+	roleImageBuilder := &builder.RoleImageBuilder{
```
f0rmiga, line 104: How about creating a `NewRoleImageBuilder`? Makes it easier to catch changes to the struct.

app/build_images.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	if opt.OutputDirectory != "" {
+		err := os.MkdirAll(opt.OutputDirectory, 0755)
+		if err != nil {
+			if os.IsExist(err) {
+				return fmt.Errorf("Output directory %s exists and is not a directory", opt.OutputDirectory)
+			}
+			if err != nil {
+				return fmt.Errorf("Error creating directory %s: %s", opt.OutputDirectory, err)
```
f0rmiga: I see sometimes you use `%s` with `err` and sometimes with `err.Error()`. Maybe we can standardize this?

app/build_images.go
```diff
@@ -0,0 +1,225 @@
...
+) error {
+
+	dockerManager, err := docker.NewImageManager()
+	if err != nil {
+		return fmt.Errorf("Error connecting to docker: %s", err.Error())
+	}
+
+	imageName, err := packagesImageBuilder.GetImageName(f.Manifest, instanceGroups, f)
+	if err != nil {
+		return fmt.Errorf("Error finding instance group's package name: %s", err.Error())
```
f0rmiga: You don't need to propagate the error message with `err.Error()`, just `fmt.Errorf("whatever meaningful: %v", err)`.

app/build_images.go
```diff
@@ -0,0 +1,225 @@
...
+		WorkerCount:        f.Options.Workers,
+	}
+
+	return roleImageBuilder.Build(instanceGroups)
+}
+
+// buildPackagesImage builds the docker image for the packages layer
+// where all packages are included
+func (f *Fissile) buildPackagesImage(
+	opt BuildImagesOptions,
```
f0rmiga, line 129: This service locator scares me a bit, especially for dealing with tests in the future.

app/build_images.go
```diff
@@ -0,0 +1,225 @@
...
+		return fmt.Errorf("Failed to find stemcell image %s. Did you pull it?", opt.Stemcell)
+	}
+
+	if opt.NoBuild {
+		f.UI.Println("Skipping packages layer docker image build because of --no-build flag.")
+		return nil
+	}
+
+	f.UI.Printf("Building packages layer docker image %s ...\n", color.YellowString(imageName))
+	log := new(bytes.Buffer)
```
f0rmiga, line 168: A buffer needs no initialization as per the documentation.

```txt
A Buffer is a variable-sized buffer of bytes with Read and Write methods. The zero value for Buffer is an empty buffer ready to use. 
```

You can simply `var log bytes.Buffer`. And when you use it, pass the pointer.

app/build_images.go
```diff
@@ -0,0 +1,225 @@
...
+
+	if !opt.Force {
+		info, err := os.Stat(outputPath)
+		if err == nil && !info.IsDir() {
+			f.UI.Printf("Packages layer %s already exists. Skipping ...\n", color.YellowString(outputPath))
+			return nil
+		}
+	}
+
+	if opt.NoBuild {
```
f0rmiga, line 204: I don't feel this check should be here. If in this method at all.

app/build_images.go
```diff
@@ -0,0 +1,225 @@
...
+		f.UI.Println("Skipping packages layer tarball build because of --no-build flag.")
+		return nil
+	}
+
+	f.UI.Printf("Building packages layer tarball %s ...\n", color.YellowString(outputPath))
+
+	tarFile, err := os.Create(outputPath)
```
f0rmiga, line 211: You want to `defer tarFile.Close()` right after the check for the err, as an error can raise at line 214, leaving the file open.

f0rmiga: Also, `os.Create` is a hidden dependency here. How will you assert for its failure on the tests?

app/build_images.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return fmt.Errorf("Failed to create tar file %s: %s", outputPath, err)
+	}
+	tarWriter := tar.NewWriter(tarFile)
+
+	// We always force build all packages here to avoid needing to talk to the
+	// docker daemon to figure out what we can keep
```
f0rmiga: [nit] Missing punctuation.

builder/release_image.go
```diff
@@ -0,0 +1,371 @@
...
+		return nil
+	})
+}
+
+// NewDockerPopulator returns a function which can populate a tar stream with the docker context to build the packages layer image with
+func (r *ReleasesImageBuilder) NewDockerPopulator(release *model.Release) func(*tar.Writer) error {
+	return func(tarWriter *tar.Writer) error {
+		// Generate dockerfile
+		dockerfile := bytes.Buffer{}
+		//if err = r.generateDockerfile(packages, labels, &dockerfile); err != nil {
```
f0rmiga: Commented code.

builder/release_image.go
```diff
@@ -0,0 +1,371 @@
...
+		err := r.generateDockerfile(&dockerfile)
+		if err != nil {
+			return err
+		}
+		err = util.WriteToTarStream(tarWriter, dockerfile.Bytes(), tar.Header{
```
f0rmiga, line 96: `WriteToTarStream` can be refactored to take the reader instead of the byte slice. Then we can use https://golang.org/pkg/io/#Copy. Perhaps a TODO?

builder/release_image.go
```diff
@@ -0,0 +1,371 @@
...
+					root:   pkg.GetPackageCompiledDir(r.CompilationDir),
+					prefix: filepath.Join("root/var/vcap/packages", pkg.Name),
+				}
+				if err = filepath.Walk(walker.root, walker.walk); err != nil {
+					return err
+				}
+			}
+
+			// Add jobs templates (including unwanted monit template)
+			err := addJobTemplates(job, "root/var/vcap/jobs-src", tarWriter)
```
f0rmiga, line 117: You may want to use `filepath.Join` for this path. Here and the other usage.

builder/release_image.go
```diff
@@ -0,0 +1,371 @@
...
+	dockerfileTemplate := template.New("Dockerfile")
+	dockerfileTemplate, err = dockerfileTemplate.Parse(string(asset))
+	if err != nil {
+		return err
+	}
+
+	return dockerfileTemplate.Execute(outputFile, context)
+}
+
+func (j releaseBuildJob) getImageName() string {
```
f0rmiga: `getImageName` -> `imageName` is more idiomatic in Go.

builder/release_image.go
```diff
@@ -0,0 +1,371 @@
...
+		if !r.Force {
+			if r.OutputDirectory == "" {
+				if hasImage, err := j.dockerManager.HasImage(imageName); err != nil {
+					return err
+				} else if hasImage {
+					r.UI.Printf("Skipping build of release image %s because it exists\n", color.YellowString(j.release.Name))
+					return nil
+				}
+			} else {
+				info, err := os.Stat(outputPath)
```
f0rmiga, line 246: `os.Stat` is a hidden dependency.

builder/release_image.go
```diff
@@ -0,0 +1,371 @@
...
+
+		if r.NoBuild {
+			r.UI.Printf("Skipping build of release image %s because of flag\n", color.YellowString(j.release.Name))
+			return nil
+		}
+
+		if r.OutputDirectory == "" {
+			r.UI.Printf("Building docker image of %s...\n", color.YellowString(j.release.Name))
+
+			log := new(bytes.Buffer)
```
f0rmiga: Same here
```go
var log bytes.Buffer
```

builder/release_image.go
```diff
@@ -0,0 +1,371 @@
...
+
+			err := j.dockerManager.BuildImageFromCallback(imageName, stdoutWriter, dockerPopulator)
+			if err != nil {
+				log.WriteTo(r.UI)
+				return fmt.Errorf("Error building image: %s", err.Error())
+			}
+		} else {
+			r.UI.Printf("Building tarball of %s...\n", color.YellowString(j.release.Name))
+
+			tarFile, err := os.Create(outputPath)
```
f0rmiga, line 297: Hidden dep.

builder/release_image.go
```diff
@@ -0,0 +1,371 @@
...
+		r.DockerOrganization = ""
+	}
+
+	dockerManager, err := newDockerImageBuilder()
+	if err != nil {
+		return fmt.Errorf("Error connecting to docker: %s", err.Error())
+	}
+
+	if r.OutputDirectory != "" {
+		if err = os.MkdirAll(r.OutputDirectory, 0755); err != nil {
```
f0rmiga, line 335: Hidden dep.

f0rmiga (changes requested):
Other than the individual comments, I miss more meaningful error propagation in general.

app/fissile_test.go
jandubois: I agree with this in principle, but the issue is pervasive throughout the project:

```
$ ag --nogroup filepath.Join.*/ | wc -l
     337
```

Given that using forward slashes work fine in all Windows APIs, I'm not going to touch this. Could be a separate cleanup PR maybe...

app/build_images.go
jandubois: Another pervasive issue. I guess at least for top-level comments that would be collected by godoc we should have trailing dots...

app/build_images.go
jandubois: I actually got rid of the `NewRoleImageBuilder` in this release. I dislike having function with 10 parameters or even more, given that Go doesn't support named parameters. Creating a struct makes the mapping of values to fields much more obvious, so you don't end up passing e.g. your booleans in the wrong order.

There was actually one field that *did* get the wrong value, but it didn't matter because the correct value was passed in later to a `Build` method, and the value from `NewRoleImageBuilder` was never used.

Anyways, I prefer this style, which is more self-documenting than using fake constructors.

app/build_images.go
jandubois: This is mostly copy&pasted. `fmt` calls `err.Error()` internally, so calling it explicitly feels kind of redundant to me, but maybe it is worth it for consistency?

app/build_images.go
jandubois: I don't know; I find 500 character error messages (because each call level added some prefix to it) rather tedious. In that case I would rather have a full stacktrace.

In this particular example, the only possible error return is already prefixed with `Error loading role manifest: %s`: https://github.com/cloudfoundry-incubator/fissile/blob/a19161d/app/fissile.go#L71-L73

I don't see how putting another prefix that is totally unrelated to the failure, but just *hints* at the call chain improves this. It is kind of obvious that the image wasn't built when the build aborted because it didn't load the manifest.

app/build_images.go
f0rmiga: I would stick to `fmt.Errorf("something: %v", err)`, which is the most widely used amongst Go folks.

app/build_images.go
jandubois: Another lost cause, unless we do a full cleanup pass:

```
$ ag --nogroup --ignore vendor 'Errorf\("[A-Z]' | wc -l
     144
```

jandubois: It's not just us though:

```
$ ag --nogroup 'Errorf\("[A-Z]' vendor | wc -l
     159
```

app/build_images.go
f0rmiga: I agree with the self-documenting. For longer initializers, I've seen and personally have used custom defined types to make it easier to track all the parameters and the propagation. I believe it is still worth it, but up to you!

jandubois: I don't understand what you are saying here, isn't `builder.RoleImageBuilder` a custom defined type? Can you point me to an example?

Looking at the previous version of the constructor, I don't see any advantage to using it over directly constructing the struct at the callers side: https://github.com/cloudfoundry-incubator/fissile/blob/a19161d/builder/role_image.go#L57-L76

How is it "worth it"?

f0rmiga:
```go
package main

import (
	"fmt"
)

type propAType string

type propBType string

type Foo struct {
	propA propAType
	propB propBType
}

func NewFoo(propA propAType, propB propBType) *Foo {
	return &Foo{
		propA: propA,
		propB: propB,
	}
}

func main() {
	pA := propAType("Hey")
	pB := propBType("there")
	foo := NewFoo(pB, pA)
	fmt.Println(foo)
}
```

This would throw the following compiler error:

```txt
prog.go:26:15: cannot use pB (type propBType) as type propAType in argument to NewFoo
prog.go:26:15: cannot use pA (type propAType) as type propBType in argument to NewFoo
```

viovanov:
Looks like @f0rmiga did the cr, I'm going to build and test locally.

jandubois:
> Looks like @f0rmiga did the cr, I'm going to build and test locally.

I've committed changes for everything from that CR that I'm willing to change (given that a lot of the comments are against code that was just moved around and not newly written).

I'll still intend to create `builder/release_image_test.go` though.

jandubois:
@viovanov I've been wondering if we should add any of the origin metadata (URL, size, SHA1) as labels on the generated images. Thoughts?

viovanov:
@jandubois finished my test run.
I think adding that metadata is a good idea.
Information about the stemcell and fissile version would also be useful.
I added these stories, including one for tests.

https://www.pivotaltracker.com/story/show/163013872
https://www.pivotaltracker.com/story/show/163013866

@f0rmiga please let us know if you think there are any blockers here.
I want to merge this, since it does what it needs to do and unblocks a bunch of work.

f0rmiga:
@viovanov If we are not adding tests to this PR, I'm happy to approve. I would just link this PR to the story so it is easier to track the changed code that needs to get tests.

f0rmiga (approved):

PR #459, opened 2019-01-07
Implement ingress support in fissile.

andreas-kupries:
Note that the majority of it is chart-specific (ingress templates, associated cert setup), and outside of the scope for fissile.

Ref: https://trello.com/c/jgev7lC4/892-3-fissile-work-to-use-an-ingress-controller
Ref: https://github.com/cloudfoundry-incubator/fissile/wiki/Managing-Ingress

### DIFF
--- a/kube/service.go
+++ b/kube/service.go
@@ -213,7 +213,7 @@ func newService(role *model.InstanceGroup, job *model.JobReference, serviceType 
 	}
 	if serviceType == newServiceTypePublic {
 		if settings.CreateHelmChart {
-			spec.Add("externalIPs", "{{ .Values.kube.external_ips | toJson }}", helm.Block("if or (not .Values.services.loadbalanced) .Values.services.ingress"))
+			spec.Add("externalIPs", "{{ .Values.kube.external_ips | toJson }}", helm.Block("if not (or .Values.services.loadbalanced .Values.services.ingress)"))
 			spec.Add("type", "LoadBalancer", helm.Block("if and .Values.services.loadbalanced (not .Values.services.ingress)"))
 		} else {
 			spec.Add("externalIPs", []string{"192.168.77.77"})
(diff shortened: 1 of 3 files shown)


jandubois (changes requested):
Please rebase this PR on top of the latest `develop`. I merged #457 earler today, which adds additional labels, that are now missing in your tests.

Or maybe it was git getting the merging wrong; I saw the "Ingress" test from `TestServiceHelm` end up in `TestIstioManagedServiceHelm` instead. So maybe double-check after the rebase that the test chunks all end up in the correct places, as some of the existing functions have been duplicated with just minor variations for Istio.

jandubois (changes requested):
I don't see the value of this change since fissile doesn't generate the ingress resource definition, so doesn't make use of the `services.ingress` name. Setting it to any non-nil name has exactly the same effect as leaving `services.loadbalanced` at the default `false` value.

I would therefore recommend to not modify fissile at all, and just add a template expression to the ingress resource that throws an error when `services.loadbalanced` is set to `true`.

jandubois:
Ok, with the updated logic for external IPs it now makes sense to have to different settings.

I think we should now include an extra check (in each "public" service):

```
{{- if and .Values.services.loadbalanced .Values.services.ingress }}
_incompatible: {{ fail "services.loadbalanced and services.ingress cannot both be set" }}
{{- end }}
```

jandubois:
> I think we should now include an extra check

And with that check we can simplify the condition:

```
spec.Add("type", "LoadBalancer", helm.Block("if and .Values.services.loadbalanced (not .Values.services.ingress)"))
```

to

```
spec.Add("type", "LoadBalancer", helm.Block("if .Values.services.loadbalanced"))
```

andreas-kupries:
I am ok with adding this new check to service templates.
And the condition change ... That is the original condition.
So, undo that part, add the config check.

jandubois (approved):

Issue #416, opened 2018-10-01
Update dependencies once https://github.com/mholt/archiver/issues/92 has landed


jandubois:
We have a local change to the vendored dependency in #414, which should be replaced by the proper fix as soon as available.

jandubois:
Upstream has been updated, so we can move to the latest archiver now. Commit f11b26d46e1610ec90dab6522c67db2cefd434ef shows source code changes required for the new API.

PR #460, opened 2019-01-09
Rework how we handle PSPs

mook-as:
We now allow the role manifest to list the PSPs / roles / cluster roles explicitly.  This way the deployment can customize what sort of (cluster) roles it wants (e.g. perhaps only one role needs to read node labels).

This should be backwards compatible: this should work with an unmodified SCF deployment.

### DIFF
--- a/app/fissile.go
+++ b/app/fissile.go
@@ -791,77 +791,95 @@ func (f *Fissile) generateAuth(settings kube.ExportSettings) error {
 		return err
 	}
 
-	// Check the accounts for the pod security policies they
-	// reference, and create their cluster roles. The necessary
-	// cluster role bindings will be created by NewRBACAccount.
-
-	for _, pspName := range model.PodSecurityPolicies() {
-		for _, accountSpec := range settings.RoleManifest.Configuration.Authorization.Accounts {
-			if accountSpec.PodSecurityPolicy == pspName {
-				node, err := kube.NewRBACClusterRolePSP(pspName, settings)
-				if err != nil {
-					return err
-				}
-				err = f.writeHelmNode(authDir, fmt.Sprintf("auth-clusterrole-%s.yaml", pspName), node)
-				if err != nil {
-					return err
-				}
-				break
-			}
-		}
-	}
-
-	for roleName, roleSpec := range settings.RoleManifest.Configuration.Authorization.Roles {
-		// Ignore roles referenced by a single instance
-		// group. These are not written as their own files,
-		// but as part of the instance group.
-		if settings.RoleManifest.Configuration.Authorization.RoleUse[roleName] < 2 {
+	// Generate accounts (and any associated role bindings / cluster role bindings)
+	for accountName, accountSpec := range settings.RoleManifest.Configuration.Authorization.Accounts {
+		// Ignore accounts referenced by a single instance group. These are not
+		// written as their own files, but as part of the instance group.
+		if len(accountSpec.UsedBy) < 2 {
 			continue
 		}
 
-		node, err := kube.NewRBACRole(roleName, roleSpec, settings)
+		nodes, err := kube.NewRBACAccount(accountName, settings.RoleManifest.Configuration, settings)
 		if err != nil {
 			return err
 		}
+		err = f.writeHelmNode(authDir, fmt.Sprintf("account-%s.yaml", accountName), nodes...)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Generate roles
+	for roleName, roleSpec := range settings.RoleManifest.Configuration.Authorization.Roles {
+		var accountNames []string
+		for accountName := range settings.RoleManifest.Configuration.Authorization.RoleUsedBy[roleName] {
+			accountNames = append(accountNames, fmt.Sprintf("- %s", accountName))
+		}
+		if len(accountNames) < 1 {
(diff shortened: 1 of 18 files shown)


jandubois (dismissed):
This PR reverts a local patch we have for `mholt/archiver`. Either that change must be excluded from the PR, or #416 must be fixed first.

mook-as:
Sorry about that; `dep` shenanigans (I added a dependency, then removed it again, leading to the kind of pointless vendor change).

Rebased to drop that commit (because it's not material to the PR).

app/fissile_test.go
```diff
@@ -417,7 +406,7 @@ func TestGenerateAuth(t *testing.T) {
 		},
 		`auth/auth-role-pointless.yaml`: []string{
 			`{
-				"apiVersion": "rbac.authorization.k8s.io/v1beta1",
+				"apiVersion": "rbac.authorization.k8s.io/v1",
```
andreas-kupries, line 409: Does this imply a minimum version of kube will be needed to handle the output ?
If yes, are we documenting this version somewhere ?

andreas-kupries:
Saw the following output when building things with the new fissile for scf, i.e. within PR https://github.com/SUSE/scf/pull/1967#pullrequestreview-191781887

```
Found 8 instance groups requiring escalation: [blobstore log-api credhub-user autoscaler-api autoscaler-servicebroker cf-usb-group routing-api api-group]
make/bosh-images
Found 8 instance groups requiring escalation: [cf-usb-group routing-api api-group blobstore log-api credhub-user autoscaler-api autoscaler-servicebroker]
```

Should I see this warning ?
And two, note the changing order of entries in the list.
Should maybe have a `sort` somewhere.
I guess this comes from a map, and Go helpfully randomizes the returned order for us.

app/fissile.go
```diff
@@ -791,77 +791,95 @@ func (f *Fissile) generateAuth(settings kube.ExportSettings) error {
...
+		sort.Strings(accountNames)
 
-		nodes, err := kube.NewRBACAccount(accountName, accountSpec, settings)
+		node, err := kube.NewRBACRole(roleName, kube.RBACRoleKindClusterRole, roleSpec, settings)
 		if err != nil {
 			return err
 		}
-		outputPath := filepath.Join(authDir, fmt.Sprintf("account-%s.yaml", accountName))
-		outputFile, err := os.Create(outputPath)
+		node.Set(helm.Comment(fmt.Sprintf(`Cluster role %s used by accounts %s`, roleName, strings.Join(accountNames, "\n"))))
```
andreas-kupries: This comes out as

```
# Cluster role privileged used by accounts default-privileged
# garden-runc
# privileged
```

I would like it better to look like

```
# Cluster role "privileged" used by accounts
# - default-privileged
# - garden-runc
# - privileged
```

I.e.: Quotes around CR name, and the list of accounts looking more like a list, with "-" prefix, and all accounts on their own line.

See also line 832 for the same for roles and accounts.

This would also apparently be more consistent with what I see for service accounts, although that uses two spaces indentation, and should also have quotes around the SA name.

```
# Service account privileged is used by the following instance groups:
#  - nfs-broker
#  - router
```

Generally all the header comments do not seem to use (double)quotes to separate object names from the text around them. IMHO it would be more readable with (double)quotes.

app/fissile.go
```diff
@@ -791,77 +791,95 @@ func (f *Fissile) generateAuth(settings kube.ExportSettings) error {
...
 			continue
 		}
+		sort.Strings(accountNames)
 
-		node, err := kube.NewRBACRole(roleName, roleSpec, settings)
+		node, err := kube.NewRBACRole(roleName, kube.RBACRoleKindRole, roleSpec, settings)
 		if err != nil {
 			return err
 		}
+		node.Set(helm.Comment(fmt.Sprintf(`Role %s used by accounts:\n%s`, roleName, strings.Join(accountNames, "\n"))))
```
andreas-kupries: Quoting issue: The `backtics` mean that the `\n` is printed as-is, instead of as newline. Use double-quotes instead. See also the comment later for related thoughts.

andreas-kupries (changes requested):

app/fissile_test.go
mook-as: We're already using `rbac.authorization.k8s.io/v1` for [`ClusterRoleBinding`](https://github.com/cloudfoundry-incubator/fissile/blob/c617cf6372fe75ebe255f2920c1dff6d6d0059ba/kube/rbac.go#L41), so this isn't actually a new requirement.  It's supported at least as far back as [Kubernetes 1.9](https://v1-9.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#clusterrolebindinglist-v1-rbac) (and docs for older versions are busted).

app/fissile.go
mook-as: Fixed in 12d0ca53.

app/fissile.go
mook-as: Fixed in 12d0ca53.

mook-as:
The requiring escalation warning has also been sorted in 12d0ca53. (Unfortunately, we don't currently have a way of returning warnings, just errors)

app/fissile_test.go
andreas-kupries: ok

andreas-kupries (approved):
Looks good to me now. Output too.

PR #461, opened 2019-01-11
Set config.use_istio to false by default in values.yaml

jandubois:
It should not matter because older role manifest should not have the "istio-managed" tag on any of their roles, but it seems cleaner to not generate the istio related labels and annotations at all unless explicitly asked for.

This is a followup to #457. After followup discussion we decided to play it safe and change the default value to maximum backwards compatibility. (fyi @zhanggbj)


f0rmiga (approved):

zhanggbj:
@jandubois 
No problem.Thanks for letting me know.

PR #462, opened 2019-01-17
Resolver: remove warning message about things needing escalation

mook-as:
Turns out we can't have any extra messages because we sometime process the output (e.g. `fissile show images`), and having extra things breaks that.


andreas-kupries (approved):
LGTM. Pity to lose the info, but correctness before nicety.

PR #465, opened 2019-01-22
Restricted kube object name sizes.

mook-as:
Kubernetes doesn't like resource names longer than 63 characters.  Try to make it so all the names are shorter than that (by including part of the hash of the full thing).

Co-authored-by: Mark Yen <mark.yen@suse.com>

### DIFF
--- a/kube/bosh_deployment_manifest_secret.go
+++ b/kube/bosh_deployment_manifest_secret.go
@@ -1,6 +1,10 @@
 package kube
 
-import "code.cloudfoundry.org/fissile/helm"
+import (
+	"fmt"
+
+	"code.cloudfoundry.org/fissile/helm"
+)
 
 // MakeBoshDeploymentManifestSecret generates a template for a secret that holds the content of a BOSH deployment manifest
 func MakeBoshDeploymentManifestSecret(settings ExportSettings) (helm.Node, error) {
@@ -9,7 +13,15 @@ func MakeBoshDeploymentManifestSecret(settings ExportSettings) (helm.Node, error
 		value = "{{ .Values.bosh | toYaml | b64enc }}"
 	}
 
-	secret := newKubeConfig(settings, "v1", "Secret", "deployment-manifest")
+	cb := NewConfigBuilder().
+		SetSettings(&settings).
+		SetAPIVersion("v1").
+		SetKind("Secret").
+		SetName("deployment-manifest")
+	secret, err := cb.Build()
+	if err != nil {
+		return nil, fmt.Errorf("failed to build a new kube config: %v", err)
+	}
 
 	data := helm.NewMapping("deployment-manifest", value)
 	secret.Add("data", data)
--- a/kube/kube_test.go
+++ b/kube/kube_test.go
@@ -95,10 +95,6 @@ func RenderNode(node helm.Node, config interface{}) ([]byte, error) {
 
 	// Note: Replicate helm's behaviour on missing keys.
 	tmpl := template.New("").Option("missingkey=zero").Funcs(functions)
-	if err != nil {
-		//fmt.Printf("TEMPLATE SETUP FAIL: %s", err)
-		return nil, err
-	}
 
 	tmpl, err = tmpl.Parse(string(helmHelpers.Bytes()))
 	if err != nil {
--- a/kube/kube_test.go
+++ b/kube/kube_test.go
@@ -68,13 +68,19 @@ func RenderNode(node helm.Node, config interface{}) ([]byte, error) {
 		return nil, fmt.Errorf("Invalid config %+v", config)
 	}
 
-	var helmConfig, yamlConfig bytes.Buffer
+	var helmConfig, yamlConfig, helmHelpers bytes.Buffer
 
 	if node == nil {
 		node = helm.NewNode(nil)
 	}
 	if err := helm.NewEncoder(&helmConfig).Encode(node); err != nil {
 		return nil, err
+	}
+
+	for _, helper := range GetHelmTemplateHelpers() {
+		if err := helm.NewEncoder(&helmHelpers).Encode(helper); err != nil {
+			return nil, err
+		}
 	}
 
 	// The functions added here are implementations of the helm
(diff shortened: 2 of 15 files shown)


kube/bosh_deployment_manifest_secret.go
```diff
@@ -9,7 +13,15 @@ func MakeBoshDeploymentManifestSecret(settings ExportSettings) (helm.Node, error
 		value = "{{ .Values.bosh | toYaml | b64enc }}"
 	}
 
-	secret := newKubeConfig(settings, "v1", "Secret", "deployment-manifest")
+	cb := NewConfigBuilder().
```
jandubois, line 16: In general I would prefer to have refactoring like this in a separate commit from the new/changed functionality. Don't bother "fixing" it for this PR; just a request for future changes.

kube/bosh_deployment_manifest_secret.go
```diff
@@ -9,7 +13,15 @@ func MakeBoshDeploymentManifestSecret(settings ExportSettings) (helm.Node, error
...
+		SetSettings(&settings).
```
jandubois, line 17: Why are `settings` passed by reference and not by value, like everywhere else? I would not expect `SetSettings` to change its argument.

kube/kube_test.go
```diff
@@ -88,12 +94,24 @@ func RenderNode(node helm.Node, config interface{}) ([]byte, error) {
 	functions["toYaml"] = renderToYaml
 
 	// Note: Replicate helm's behaviour on missing keys.
-	tmpl, err := template.New("").Option("missingkey=zero").Funcs(functions).Parse(string(helmConfig.Bytes()))
+	tmpl := template.New("").Option("missingkey=zero").Funcs(functions)
+	if err != nil {
```
jandubois: Where is `err` being set here?

kube/rbac.go
```diff
@@ -48,12 +48,19 @@ func NewRBACAccount(accountName string, config *model.Configuration, settings Ex
...
-			helm.Comment(description)))
+
+		cb := NewConfigBuilder().
+			SetSettings(&settings).
+			SetAPIVersion("v1").
+			SetKind("ServiceAccount").
+			SetName(accountName).
+			AddModifier(block).
+			AddModifier(helm.Comment(description))
+		resource, err := cb.Build()
```
jandubois: Why not call it `serviceAccount` instead of `resource`? That leads to the more readable

```
resources = append(resources, serviceAccount)
```

instead of the eye-squinting

```
resources = append(resources, resource)
```

kube/rbac.go
```diff
@@ -122,23 +134,39 @@ func NewRBACAccount(accountName string, config *model.Configuration, settings Ex
...
-			"rbac.authorization.k8s.io/v1",
-			"ClusterRoleBinding",
-			authCRBindingName(accountName, clusterRoleName, settings),
-			block,
-			helm.Comment(fmt.Sprintf(`Cluster role binding for service account "%s" and cluster role "%s"`,
+		cb := NewConfigBuilder().
+			SetSettings(&settings).
+			SetAPIVersion("rbac.authorization.k8s.io/v1").
+			SetKind("ClusterRoleBinding").
+			SetName(fmt.Sprintf("%s-%s-cluster-binding", accountName, clusterRoleName)).
```
jandubois: Don't care too much, but maybe set the `Name` in the `else` branch of `if settings.CreateHelmChart` below, for symmetry with how you set the `roleRef` name a little further down?

kube/templates.go
```diff
@@ -0,0 +1,29 @@
...
+			{{ if lt (len .) 1 }}{{ fail "No name given for node" }}{{ end }}
+			{{ if gt (len .) 63 }}
+				{{ . | trunc 54 }}-{{ . | sha256sum | trunc 8 }}
+			{{ else }}
+				{{ . }}
+			{{ end }}
+		{{ end }}`
+	return []helm.Node{
+		helm.NewNode(
+			regexp.MustCompile(`(^|\}\})\s+(\{\{|$)`).
```
jandubois: Is this really necessary? I just spent 3 minutes figuring out what the regexp is supposed to do and to convince myself that is correct and not doing too much or too little. At a minimum it should have a comment on top.

It would be much simpler/obvious to use the templating mechanism for eating unwanted whitespace: `{{- expre -}}`.

kube/utils.go
```diff
@@ -39,9 +39,84 @@ func newSelector(role *model.InstanceGroup, settings ExportSettings) *helm.Mappi
 	return meta
 }
 
-// newKubeConfig sets up generic a Kube config structure with minimal metadata
-func newKubeConfig(settings ExportSettings, apiVersion, kind, name string, modifiers ...helm.NodeModifier) *helm.Mapping {
-	labels := helm.NewMapping(RoleNameLabel, name) // "app.kubernetes.io/component"
+// ConfigBuilder sets up generic a Kube resource structure with minimal metadata.
```
jandubois:
```suggestion
// ConfigBuilder sets up a generic Kube resource structure with minimal metadata.
```

kube/utils.go
```diff
@@ -39,9 +39,84 @@ func newSelector(role *model.InstanceGroup, settings ExportSettings) *helm.Mappi
...
+// NewConfigBuilder constructs a new ConfigBuilder.
+func NewConfigBuilder() *ConfigBuilder {
+	return &ConfigBuilder{}
+}
+
+func (b *ConfigBuilder) setError(err error) {
+	if b.err == nil {
+		b.err = err
+	} else {
+		b.err = fmt.Errorf("%v: %v", b.err, err)
```
jandubois: Why `:` instead of `,`; the errors can be independent of each other. The `:` implies that the right side provides more details about the left.

Nit-pickers-r-us

kube/utils.go
```diff
@@ -39,9 +39,84 @@ func newSelector(role *model.InstanceGroup, settings ExportSettings) *helm.Mappi
...
+}
+
+// SetNameHelmExpression sets the name of the resource to build as a Helm template expression.
+func (b *ConfigBuilder) SetNameHelmExpression(name string) *ConfigBuilder {
+	if b.settings == nil {
+		b.setError(fmt.Errorf("name was set as a Helm expression before settings was set"))
+	} else if !b.settings.CreateHelmChart {
+		b.setError(fmt.Errorf("name is a Helm expression, but not creating a helm chart"))
+	}
+	b.name = name
```
jandubois, line 109: You could add a check here that `name` starts with `{{`. Otherwise `helm/config.go` might put quotes around the expression. Not sure if it matters in this specific case.

jandubois (commented):
None of the comments totally require changes, but at least the unset `err` thing should be fixed, I think. (It is going to work as-is, because `err` is always `nil` at that point, but then the whole condition doesn't make sense).

kube/kube_test.go
mook-as: It's not; my mistake. Sorry about that.

jandubois (approved):

kube/utils.go
f0rmiga: Thanks for pointing this out. You are absolutely right, my bad.

kube/bosh_deployment_manifest_secret.go
f0rmiga: Passing the reference doesn't mean it's going to be changed.

PR #470, opened 2019-02-14
Add support for reverse consumer information

mook-as:
In `/var/vcap/job-src/*/config_spec.json`, we will now include a list of consumers of this job's links (via the `consumed_by` key, similar to the existing `consumes` key).  This is useful so we can attempt to restart pods which contain outdated information (in case they linked against a previous iteration of this pod and now contains outdated information).

/cc @viovanov


model/job_reference.go
```diff
@@ -11,7 +11,8 @@ type JobReference struct {
 	Name                string                     `yaml:"name"`    // The name of the job
 	ReleaseName         string                     `yaml:"release"` // The release the job comes from
 	ExportedProviders   map[string]JobProvidesInfo `yaml:"provides"`
-	ResolvedConsumers   map[string]JobConsumesInfo `yaml:"consumes"`
+	ResolvedConsumers   map[string]JobConsumesInfo `yaml:"consumes"`    // Instance groups that this job links to & requires
```
f0rmiga, line 14: How about aligning the struct field label with the yaml field label? It is a bit confusing one being `...Consumers` and the other `consumes`.

model/resolver/resolver.go
```diff
@@ -345,6 +345,43 @@ func (r *Resolver) ResolveLinks() validation.ErrorList {
 		}
 	}
 
+	// Record in each job what roles consume it
```
f0rmiga: [nit] Punctuation here and below.

model/resolver/resolver.go
```diff
@@ -345,6 +345,43 @@ func (r *Resolver) ResolveLinks() validation.ErrorList {
...
+	for _, consumerInstanceGroup := range m.InstanceGroups {
```
f0rmiga, line 372: What do you think about pulling the logic from the inner-most loop into a function?

model/resolver/resolver.go
```diff
@@ -345,6 +345,43 @@ func (r *Resolver) ResolveLinks() validation.ErrorList {
...
+		for _, consumerJob := range consumerInstanceGroup.JobReferences {
+			for linkName, consumer := range consumerJob.ResolvedConsumers {
+				providerInstanceGroup := m.LookupInstanceGroup(consumer.RoleName)
+				if providerInstanceGroup == nil {
+					// This should not happen: we resolved a link, but can no
+					// longer find the instance group that provides it
+					errors = append(errors, validation.InternalError(
+						fmt.Sprintf(`instance_group[%s].job[%s].consumes[%s]`, consumerInstanceGroup.Name, consumerJob.Name, linkName),
```
f0rmiga: Declaring variables in a more procedural approach would help reading it. E.g.:
```go
field := fmt.Sprintf("instance_group[%s].job[%s].consumes[%s]", consumerInstanceGroup.Name, consumerJob.Name, linkName)
err := fmt.Errorf("Could not find resolved instance group %s", consumer.RoleName)
internalErr := validation.InternalError(field, err)
errors = append(errors, internalErr)
```

Also, no need for using ` for the strings.

model/resolver/resolver.go
```diff
@@ -345,6 +345,43 @@ func (r *Resolver) ResolveLinks() validation.ErrorList {
...
+				providerJob := providerInstanceGroup.LookupJob(consumer.JobName)
+				if providerJob == nil {
+					// This should not happen: we resolved a link, but can no
+					// longer find the job that provides it
+					errors = append(errors, validation.InternalError(
+						fmt.Sprintf(`instance_group[%s].job[%s].consumes[%s]`, consumerInstanceGroup.Name, consumerJob.Name, linkName),
+						fmt.Errorf(`Could not find resolved job %s in instance group %s`, consumer.JobName, consumer.RoleName)))
+					continue
+				}
+				if providerJob.ResolvedConsumedBy == nil {
```
f0rmiga: I think it should be initialized when the object is created.

model/resolver/validation.go
```diff
@@ -54,6 +54,10 @@ func validateInstanceGroup(roleManifest *model.RoleManifest, g *model.InstanceGr
 			info.Name = name
 			jobReference.ResolvedConsumers[name] = info
 		}
+
+		if jobReference.ResolvedConsumedBy == nil {
```
f0rmiga, line 58: Why a validation function is mutating a parameter passed to it? I see the previous code is doing it already, but it is way out of its concern doing so. Could you add a TODO for moving it out, and create a card for it on our board?

f0rmiga (changes requested):

model/job_reference.go
mook-as: I feel like that should be done separately (this was just a comment change; refactoring would be a bigger change that should get its own PR).  I might still be convinced though!

f0rmiga: It is ok to do in a separate PR, so we don't mix things up! Could we do right after?

model/resolver/validation.go
f0rmiga: Added a card to tackle this.

model/job_reference.go
f0rmiga: Card created to tackle this.

f0rmiga (approved):

PR #471, opened 2019-02-15
Add fissile and stemcell version to the release image tags

aduffeck:


jandubois (changes requested):
I haven't looked into the changes themselves, but a simple test that works with the `develop` branch fails with this PR:

```
$ fissile build release-images --repository test --stemcell splatform/fissile-stemcell-opensuse --name nats --sha1 2f2c27acdfff81f3519968921686522518ab5783 --url 'https://bosh.io/d/github.com/cloudfoundry/nats-release?v=26' --version 26
Error looking up image test-nats:--7.0.0_234.gcd7d113-26: API error (400): no such image: test-nats:--7.0.0_234.gcd7d113-26: invalid reference format
```

aduffeck:
@jandubois With this change fissile depends on the new stemcells for calculating the release image tag which include labels for the stemcell flavor and version. The stemcell you used supposedly doesn't have these labels yet.

I'll improve the error reporting for such cases tomorrow, though.

aduffeck:
As per a discussion with @viovanov  we'll show a warning and resort to `unknown-<image-id>` instead of erroring out when the stemcell image doesn't have the labels.

jandubois (approved):
I don't exactly love the format of the new tags:

```
opensuse-42.3-28.g837c5b3-30.75-7.0.0_235.ga652925-26
```

It doesn't look obvious how to split it into the 3 separate parts automatically, but I guess that isn't really needed.

jandubois:
@aduffeck PRs should include a reference to the tracker id, so that bots can add a reference automatically.

[#163916819]

jandubois:
Hmm, I guess it only works in the main description field, or in the commit messages...

aduffeck:
@jandubois Yes, tracker is really picky about the format and the location of the reference. iirc only the pr description worked when we played around with it.
I didn't put a reference here because I thought that the fissile repo wasn't connected to the cfcontainerization tracker, but maybe I'm mistaken. I'll give it a try next time.

jandubois:
> I didn't put a reference here because I thought that the fissile repo wasn't connected to the cfcontainerization tracker

@aduffeck It does seem to be hooked up; here is an auto-added commit reference: https://www.pivotaltracker.com/n/projects/2192232/stories/163013866

PR #472, opened 2019-02-19
No `*-set` service for individual jobs

viovanov:
Testing if this is a fix for some service name collisions.
This doesn't look like it's needed - since there's the umbrella headless service that exposes all ports from all jobs.

DON'T MERGE


viovanov:
closing in favor of #473

PR #473, opened 2019-02-19
Fix duplicate headless services 

qu1queee:
Current setup will generate two headless svc´s, these are `<role>-<job>-set` and `<role>-set`.
When using the `service_name` key in an specific job, in order to rename the svc´s, having both headless svc´s will lead to duplicate svc´s .

A more precise example is the `nats` role. Without using the `service_name` key in an `instance_group` job, the headless svcs will be:
```
nats-nats-set
nats-set
```
When using the  `service_name` key, to enforce the name of `uaa`. the headless svcs will be:
```
nats-set
nats-set
```
which will lead to the svc´s duplication.

After speaking with @viovanov about this, he mentioned that the the `<role>-<job>-set` svc, is not widely used, and only a reference to it is found in:
 - https://github.com/SUSE/scf/blob/develop/container-host-files/etc/scf/config/role-manifest.yml#L2241

Therefore this PR to remove the `<role>-<job>-set` svc .

### DIFF
--- a/kube/service.go
+++ b/kube/service.go
@@ -25,17 +25,6 @@ func NewServiceList(role *model.InstanceGroup, clustering bool, settings ExportS
 	}
 
 	for _, job := range role.JobReferences {
-		if clustering {
-			// Create headless, private service
-			svc, err := newService(role, job, newServiceTypeHeadless, settings)
-			if err != nil {
-				return nil, err
-			}
-			if svc != nil {
-				items = append(items, svc)
-			}
-		}
-
 		// Create private service
 		svc, err := newService(role, job, newServiceTypePrivate, settings)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


jandubois:
I would like to see a CI build of SCF that passes tests with this change to fissile before merging it into `develop`.

qu1queee:
This will not work when renaming the services for `nats`. Reason is the following line: https://github.com/SUSE/scf/blob/2.14.5/container-host-files/etc/scf/config/scripts/configure-HA-hosts.sh#L41

jandubois:
Yes, this will require additional work. 

This name conflict is the reason we renamed some roles, e.g. `api` to `api-group`, to make the original service name `api` available again as an alternate service name for backwards compatibility. Same reason for renaming `diego-access` to `diego-ssh`.

viovanov:
The `-set` is still not required.
Renaming for nats may not work but it's not because of this.
This should be merged afaict.

PR #477, opened 2019-03-14
Replace combination of build-time smylinks and runtime moves with build time moves.

andreas-kupries:
Ref https://trello.com/c/H2AXtiHz/974-packages-mv-should-happen-at-build-time

Controlling SCF PR https://github.com/SUSE/scf/pull/2191

Drop symlinks for the role tarball.
Use `mv` in the Dockerfile instead.
Removed transform of symlinks into `mv` from the runtime role startup.

:white_check_mark: check unit tests
:white_check_mark: Get check build
:white_check_mark: Trial with SCF (and actual packages to move)

### DIFF
--- a/builder/role_image.go
+++ b/builder/role_image.go
@@ -88,29 +88,6 @@ func (r *RoleImageBuilder) NewDockerPopulator(instanceGroup *model.InstanceGroup
 			}
 		}
 
-		// Symlink compiled packages
-		packageSet := map[string]string{}
-		for _, jobReference := range instanceGroup.JobReferences {
-			for _, pkg := range jobReference.Packages {
-				if _, ok := packageSet[pkg.Name]; !ok {
-					err := util.WriteToTarStream(tarWriter, nil, tar.Header{
-						Name:     filepath.Join("root/var/vcap/packages", pkg.Name),
-						Typeflag: tar.TypeSymlink,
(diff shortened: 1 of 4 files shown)


builder/role_image.go
```diff
@@ -300,14 +277,42 @@ func (r *RoleImageBuilder) generateDockerfile(instanceGroup *model.InstanceGroup
...
 	}
 
-	dockerfileTemplate := template.New("Dockerfile-role")
+	// In a role move the packages from /var/vcap/packages-src to
+	// /var/vcap/packages.  This replaces the original form of
+	// symlinking them into place.  The change is necessary
+	// because the symlinks will not resolve inside bpm containers.
+
+	packageSet := map[string]string{}
+	moves := []string{}
```
jandubois: I don't like the name `moves`; these are really `commands` of any kind.

I can't make up my mind, but I think I might even prefer to have the `RUN` prefix in the template, and just add regular shell commands to a `runCommands` array that you expand with

```
{{ range .runCommands }}
RUN {{ . }}
{{ end }}
```

andreas-kupries: Made this change, plus a few other fixes: Typo in the paths, and forgot to create destination directory.
Ref commit 6a73892c23d0eac50861ac6c031012d338d90a80

jandubois (approved):

PR #480, opened 2019-03-18
Fix scaling default handling

jandubois:


jandubois:
This PR has been obsoleted by the feature flag changes (which reverted the introduction of Scaling.Default).

PR #482, opened 2019-03-28
Add support for feature flags to role manifests

jandubois:


andreas-kupries (approved):
Looks good to me. Product decided to go with this instead of the change for supporting default counts. (Undone here).

PR #483, opened 2019-03-28
nproc limit

andreas-kupries:
Ref https://trello.com/c/XvU6E5Vg/955-allow-bumping-up-the-nproc-limit-for-vcap-user

Controlling PR https://github.com/SUSE/scf/pull/2244

### DIFF
--- a/kube/pod.go
+++ b/kube/pod.go
@@ -508,12 +508,21 @@ func getEnvVarsFromConfigs(configs model.Variables, settings ExportSettings) (he
 		env = append(env, helm.NewMapping("name", config.Name, "value", stringifiedValue))
 	}
 
-	fieldRef := helm.NewMapping("fieldPath", "metadata.namespace")
+	{
+		fieldRef := helm.NewMapping("fieldPath", "metadata.namespace")
+		envVar := helm.NewMapping("name", "KUBERNETES_NAMESPACE")
+		envVar.Add("valueFrom", helm.NewMapping("fieldRef", fieldRef))
+		env = append(env, envVar)
+	}
 
-	envVar := helm.NewMapping("name", "KUBERNETES_NAMESPACE")
-	envVar.Add("valueFrom", helm.NewMapping("fieldRef", fieldRef))
-
-	env = append(env, envVar)
+	{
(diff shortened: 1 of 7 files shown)


kube/pod_test.go
```diff
@@ -580,13 +600,16 @@ func TestPodGetEnvVarsFromConfigGenerationCounterHelm(t *testing.T) {
...
 	}
 	testhelpers.IsYAMLEqualString(assert, `---
-		-	name: "KUBERNETES_NAMESPACE"
-			valueFrom:
-				fieldRef:
-					fieldPath: "metadata.namespace"
-		-	name: "KUBE_SECRETS_GENERATION_COUNTER"
-			value: "3"
-	`, actual)
+- name: "KUBERNETES_NAMESPACE"
```
f0rmiga: I had an issue with indentation here. I didn't have time to get back to it before pushing the WIP commit (by the way, I'll rename that commit). I'll push a commit to fix it up.

kube/pod.go
```diff
@@ -509,12 +509,28 @@ func getEnvVarsFromConfigs(configs model.Variables, settings ExportSettings) (he
 	}
 
 	fieldRef := helm.NewMapping("fieldPath", "metadata.namespace")
-
 	envVar := helm.NewMapping("name", "KUBERNETES_NAMESPACE")
 	envVar.Add("valueFrom", helm.NewMapping("fieldRef", fieldRef))
-
 	env = append(env, envVar)
 
+	if settings.CreateHelmChart {
```
mook-as, line 516: Wouldn't this magic make more sense living near the `KUBE_SECRETS_GENERATION_COUNTER` &c stuff above, to keep it all in the same area?

mook-as (changes requested):

kube/pod.go
andreas-kupries: Checking ...

andreas-kupries: @mook-as The `COUNTER` variable, etc. is in a loop.
The `NPROC` setup happens after that loop.
These are definitely distinct areas.
I don't believe I can merge this, or that it would make sense to move `NPROC` into the loop. They are not part of the `configs` the loop iterates over.

IMHO nothing to do.

mook-as (approved):

PR #484, opened 2019-03-29
Fix indentation consistency

f0rmiga:
## Description

- `run.sh` had tabs mixed up with spaces. Preferred spaces over tabs as spaces are the majority.
- Used 2-space indentation when it was 4-space.
- Fixed comment punctuation.


mook-as (approved):
I still like indenting the source (as opposed to matching the indent for the generated file), but not enough to not merge this.

Please add a modeline on top so we know the canonical indent size.

PR #485, opened 2019-04-01
Replaced pre-start scripts running logic using Go templating

f0rmiga:
## Description

Replaced the `find` approach with a Go templating approach. It executes the pre-start scripts in the order that was passed to the jobs field in the role manifest.

## Test plan

Make sure that the order passed to the jobs is respected when rendering `run.sh`.


andreas-kupries (approved):
Reviewed. Looks ok. Tested on vagrant, startup is good.
Jenkins is irrelevant because SCF PR did not reference the changed fissile.

PR #486, opened 2019-04-02
Invoke pre-start scripts via explicit `bash`

andreas-kupries:
Ref https://trello.com/c/Ljs7a8lx/986-2-make-sure-authorize-internal-ca-pre-start-script-always-runs-first

PS to https://github.com/cloudfoundry-incubator/fissile/pull/485
Invoke the scripts with an explicit `bash`, like the old code did.

It looks as if at least one pre-start script has active options when called directly.

This PR is for testing if that is indeed the case.
The "proper" change would be to fix the pre-start script, given that it is the only one (so far).
This change here is a bit easier to create

:construction: :warning: :construction: 
__Do not merge__


andreas-kupries:
Closing, confirmed issue, fix in SCF itself, keeping fissile as-is

PR #489, opened 2019-04-03
Fix colocated container volume count validation

### LINKED ISSUE #488: Role manifest shared volume validation bug
We ran into an issue where the role manifest validation failed for our setup, where one instance group has multiple colocated container instance groups. Not all of the colocated instance groups have an `emptyDir` volume defined.

However, the validation we initially wrote for fissile did not take that scenario into account and simply checked that the number of volumes match. This does not work for the aforementioned setup. The number of colocated containers is two, but the number of volumes is one. Therefore, the validation fails.

```yaml
instance_groups:
- name: main-instance-group
  jobs:
  - name: tor
    release: tor
    properties:
      bosh_containerization:
        colocated_containers:
        - to-be-colocated-1
        - to-be-colocated-2
        run:
          memory: 1
          volumes:
          - path: /shared/data
            type: emptyDir
            tag: shared-data

- name: to-be-colocated-1
  type: colocated-container
  jobs:
  - name: tor
    release: tor
    properties:
      bosh_containerization:
        run:
          memory: 1
          volumes:
          - path: /shared/data
            type: emptyDir
            tag: shared-data

- name: to-be-colocated-2
  type: colocated-container
  jobs:
  - name: ntpd
    release: ntp
    properties:
      bosh_containerization:
        # no volume defined here ...
        run:
          memory: 1
```

HeavyWombat:
Fix the colocated container volume validation that checks for unsed volumes
defined in the main instance group, which are not used by any colocated
container instance group. Do not compare the number of volumes, but check
that each volume definition of the main instance group is at least used
once by one of the colocated containers. The previous logic created false
validation errors in case multiple colocated containers were defined, but
not all of them used a shared volume.

This should fix #488.

### DIFF
--- a/model/resolver/validation.go
+++ b/model/resolver/validation.go
@@ -529,9 +529,7 @@ func validateColocatedContainerVolumeShares(roleManifest *model.RoleManifest) va
 	allErrs := validation.ErrorList{}
 
 	for _, instanceGroup := range roleManifest.InstanceGroups {
-		numberOfColocatedContainers := len(instanceGroup.ColocatedContainers())
-
-		if numberOfColocatedContainers > 0 {
+		if len(instanceGroup.ColocatedContainers()) > 0 {
 			emptyDirVolumesTags := []string{}
 			emptyDirVolumesPath := map[string]string{}
 			emptyDirVolumesCount := map[string]int{}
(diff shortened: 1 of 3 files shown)


model/resolver/validation.go
```diff
@@ -541,37 +539,46 @@ func validateColocatedContainerVolumeShares(roleManifest *model.RoleManifest) va
...
-							}
+						// Validate that volumes with the same tag in both the main instance group
+						// and the colocated container have the same path configured. The same tag
+						// but a different path results in a validation error.
+						if path, ok := emptyDirVolumesPath[volume.Tag]; ok && path != volume.Path {
+							// Same tag, but different paths
+							allErrs = append(allErrs, validation.Invalid(
+								fmt.Sprintf("instance_group[%s]", colocatedInstanceGroup.Name),
+								volume.Path,
+								fmt.Sprintf("colocated instance group specifies a shared volume with tag %s, which path does not match the path of the main instance group shared volume with the same tag", volume.Tag)))
```
mook-as, line 568: Not really part of your PR, but: shouldn't `instanceGroup.Name` be shown somewhere?

mook-as (approved):
PR looks fine, seems reasonable.

mook-as:
Test run passed locally (and correctly failed when I reverted the change locally); merging.

PR #490, opened 2019-04-08
Add feature flag details to values.yaml comments

jandubois:
There are *no* functional changes; all changes only affect the comments inside `values.yaml`.

For each instance group that has a feature flag condition this change adds a comment to the `sizing.INSTANCEGROUP.count` field, e.g.

```
sizing:
  diego_api:
    ...
    # The diego_api instance group can be disabled by the eirini feature.
    # It can scale between 1 and 3 instances.
    # For high availability it needs at least 2 instances.
    count: 1
```

Each feature flag is also annotated with the list of instance groups it controls, e.g.

```
enable:
  # The eirini feature enables these instance groups: eirini
  # It disables these instance groups: diego_api, diego_brain, diego_ssh, and
  # diego_cell
  eirini: false
```

Additional minor changes are adding the conjunctive "and" in the list of 2 or more jobs, and using the proper format for the instance group names (underscores instead of dashes, `credhub_user` instead of `credhub-user`).

### DIFF
--- a/kube/values.go
+++ b/kube/values.go
@@ -6,6 +6,7 @@ import (
 
 	"code.cloudfoundry.org/fissile/helm"
 	"code.cloudfoundry.org/fissile/model"
+	"code.cloudfoundry.org/fissile/util"
 )
 
 func formattedExample(example string) string {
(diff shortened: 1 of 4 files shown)


mook-as (approved):
Looks like it should do the reported thing.

This looks like it makes fissile harder to hack on though (because the code is getting more complicated), and especially `MakeValues()` should probably be split up some.  Also there are no tests for the new comments (thanks for the test on `WordList`, of course).

jandubois:
> This looks like it makes fissile harder to hack on though

Yes, slightly, but isn't that true for all added functionality. I thought it is worth it because it provides the end user with additional documentation in the generated `values.yaml`. All the code for the autogenerated docs is slightly complicated for what it does, to keep things somewhat readable. And it still does some things wrong, like using plural forms when only giving a single value in a list. But I thought fixing *that* would not be worth the additional complexity.

> `MakeValues()` should probably be split up some

Agreed, but not in this PR.

> Also there are no tests for the new comments

Agreed again. The problem is that there aren't any tests for the existing comments either, so adding tests for all of them would be quite a big task now. If we want to do that, then we should have a separate card for it.

Not sure it is worth it, given that SCF will no longer use the generated helm charts once it switches to using cf-operator.

PR #491, opened 2019-04-10
Ingress default values

f0rmiga:
## Description

- Added ingress default values with comments.
- Updated values_test.go.

## Test plan

Fissile should generate the `values.yaml` with the specified ingress default values (as seen on the unit test).

### DIFF
--- a/kube/values.go
+++ b/kube/values.go
@@ -266,7 +266,7 @@ func MakeValues(settings ExportSettings) helm.Node {
 	values.Add("enable", enable.Sort())
 
 	ingress := helm.NewMapping()
-	ingress.Add("annotations", helm.NewMapping(), helm.Comment("ingress.annotations specify custom ingress annotations that gets merged to the default annotations."))
+	ingress.Add("annotations", helm.NewMapping(), helm.Comment("ingress.annotations allows specifying custom ingress annotations that gets merged to the default annotations."))
 	ingress.Add("enabled", false, helm.Comment("ingress.enabled enables ingress support - working ingress controller necessary."))
 	ingress.Add("tls", helm.NewMapping(), helm.Comment("ingress.tls.crt and ingress.tls.key, when specified, are used by the TLS secret for the Ingress resource."))
 	values.Add("ingress", ingress.Sort())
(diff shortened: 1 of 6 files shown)


kube/values.go
```diff
@@ -265,5 +265,11 @@ func MakeValues(settings ExportSettings) (helm.Node, error) {
 	}
 	values.Add("enable", enable.Sort())
 
-	return values, nil
+	ingress := helm.NewMapping()
+	ingress.Add("annotations", helm.NewMapping(), helm.Comment("ingress.annotations allows specifying custom ingress annotations that gets merged to the default annotations."))
+	ingress.Add("enabled", false, helm.Comment("ingress.enabled enables ingress support - working ingress controller necessary."))
+	ingress.Add("tls", helm.NewMapping(), helm.Comment("ingress.tls.crt and ingress.tls.key, when specified, are used by the TLS secret for the Ingress resource."))
+	values.Add("ingress", ingress.Sort())
```
mook-as, line 272: Testing out inlining this…

```golang
	values.Add("ingress", helm.NewMapping(
		"annotations", helm.NewNode(map[string]interface{}{}, helm.Comment("ingress.annotations allows specifying custom ingress annotations that gets merged to the default annotations.")),
		"enabled", helm.NewNode(false, helm.Comment("ingress.enabled enables ingress support - working ingress controller necessary.")),
		"tls", helm.NewNode(map[string]interface{}{}, helm.Comment("ingress.tls.crt and ingress.tls.key, when specified, are used by the TLS secret for the Ingress resource.")),
	).Sort())
```

Hmm, not sure I like one better than the other.

mook-as (approved):
Generally okay, probably better to split out unrelated cleanup though :(

kube/values.go
f0rmiga: I rarely prefer this type of reduction. In this case I'm ok with either.

mook-as (changes requested):
https://github.com/cloudfoundry-incubator/fissile/blob/e67442aada507ce1b9e6c45bef44dec6dca876bd/kube/utils.go#L219 needs to remove `services.ingress` (or possibly move the whole new part in there).

mook-as (approved):

PR #492, opened 2019-04-16
Replace deprecated storage-class annotation with storageClassName key.

andreas-kupries:
Ref: https://trello.com/c/rhJZ3Rqm/1036-dont-use-deprecated-storage-class-key-in-helm-templates

### DIFF
--- a/kube/stateful_set.go
+++ b/kube/stateful_set.go
@@ -87,11 +87,7 @@ func getVolumeClaims(role *model.InstanceGroup, createHelmChart bool) []helm.Nod
 
 		meta := helm.NewMapping("name", volume.Tag)
 		if len(volume.Annotations) > 0 {
-			annotationList := helm.NewMapping()
-			for key, value := range volume.Annotations {
-				annotationList.Add(key, value)
-			}
-			meta.Add("annotations", annotationList)
+			meta.Add("annotations", helm.NewNode(volume.Annotations))
 		}
 
 		var size string
--- a/kube/stateful_set.go
+++ b/kube/stateful_set.go
@@ -88,22 +88,10 @@ func getVolumeClaims(role *model.InstanceGroup, createHelmChart bool) []helm.Nod
 		meta := helm.NewMapping("name", volume.Tag)
 		if len(volume.Annotations) > 0 {
 			annotationList := helm.NewMapping()
-			haveAnnotations := false
 			for key, value := range volume.Annotations {
-				// Detect and convert a storage-class annotation in the role-manifest to
-				// a regular storage class name. Needed because such an annotation is the
-				// only way for the role manifest to override the type-derived storage
-				// class, yet we must not emit this specific annotation anymore.
-				if key == VolumeStorageClassAnnotation {
-					storageClass = value
-					continue
-				}
 				annotationList.Add(key, value)
-				haveAnnotations = true
 			}
-			if haveAnnotations {
-				meta.Add("annotations", annotationList)
-			}
+			meta.Add("annotations", annotationList)
 		}
 
 		var size string
--- a/kube/stateful_set.go
+++ b/kube/stateful_set.go
@@ -86,12 +86,25 @@ func getVolumeClaims(role *model.InstanceGroup, createHelmChart bool) []helm.Nod
 		}
 
 		meta := helm.NewMapping("name", volume.Tag)
-		annotationList := helm.NewMapping()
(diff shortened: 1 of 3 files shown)


kube/stateful_set.go
```diff
@@ -86,12 +86,25 @@ func getVolumeClaims(role *model.InstanceGroup, createHelmChart bool) []helm.Nod
...
 
 		meta := helm.NewMapping("name", volume.Tag)
-		annotationList := helm.NewMapping()
-		annotationList.Add(VolumeStorageClassAnnotation, storageClass)
-		for key, value := range volume.Annotations {
-			annotationList.Add(key, value)
+		if len(volume.Annotations) > 0 {
+			annotationList := helm.NewMapping()
+			haveAnnotations := false
+			for key, value := range volume.Annotations {
```
mook-as: Given that `volume.Annotations` is a `map[string]string`, wouldn't it be easier to write this as something like:

```golang
annoationList := helm.NewMapping(volume.Annotations)
if value, ok := volume.Annotations[VolumeStorageClassAnnotation]; ok {
	storageClass = value
	annotationList.Delete(VolumeStorageClassAnnotation)
}
```
Will need to implement `Mapping.Delete` though.  Possibly something to check for emptiness if we don't want to emit empty annotations too.

kube/stateful_set.go
```diff
@@ -86,12 +86,25 @@ func getVolumeClaims(role *model.InstanceGroup, createHelmChart bool) []helm.Nod
...
+				// Detect and convert a storage-class annotation in the role-manifest to
+				// a regular storage class name. Needed because such an annotation is the
+				// only way for the role manifest to override the type-derived storage
+				// class, yet we must not emit this specific annotation anymore.
```
jandubois: Are we actually using this anywhere?

andreas-kupries: I saw this in a unit test, and not having this code breaks that test. As such I assumed that this is something which can happen in the wild. 

See
`./test-assets/role-manifests/kube/volumes-with-annotation.yml:21:              volume.beta.kubernetes.io/storage-class: a-company-file-gold`

jandubois: "In the wild" essentially means in SCF or UAA, and I don't see them using it.

I would prefer to get rid of this "feature" instead of writing new code to implement it. If you decide to keep this code, then I think you also need to figure out what to do with the `storage-provisioner` annotation:

```
  annotations:
    volume.beta.kubernetes.io/storage-class: a-company-file-gold
    volume.beta.kubernetes.io/storage-provisioner: a-company.io/storage-provisioner
```

It seems inconsistent to transform one, but not the other. Not translating either of them seems easier. 😄

andreas-kupries: Ok. Will remove that part, and update the unit test to match.

kube/stateful_set.go
andreas-kupries: Looks to become moot with my decision to remove this translation.

kube/stateful_set.go
andreas-kupries: Addressed with commit 0662aedf2624e45f8529df6d0c250d8196fe7902.

kube/stateful_set.go
andreas-kupries: See commit 0662aedf2624e45f8529df6d0c250d8196fe7902.

andreas-kupries:
Rebased to get the ingress changes.

kube/stateful_set.go
```diff
@@ -86,12 +86,13 @@ func getVolumeClaims(role *model.InstanceGroup, createHelmChart bool) []helm.Nod
 		}
 
 		meta := helm.NewMapping("name", volume.Tag)
-		annotationList := helm.NewMapping()
-		annotationList.Add(VolumeStorageClassAnnotation, storageClass)
-		for key, value := range volume.Annotations {
-			annotationList.Add(key, value)
+		if len(volume.Annotations) > 0 {
```
jandubois, line 89: Aaron always hated it, but the `helm` code uses reflection to dynamically deal with data types, so you can simplify this as:

```
		if len(volume.Annotations) > 0 {
			meta.Add("annotations", helm.NewNode(volume.Annotations))
		}
```

andreas-kupries: Ok.

andreas-kupries: Done with commit 80d42a64006c290843a576942e286a7911df71e4

jandubois (approved):

PR #494, opened 2019-04-22
Don't generate imagePullSecrets when the username is empty

### LINKED ISSUE #487: Deployment fails on clusters with containerd when credentials are empty
This line does not work when we don't need credentials for Docker hub
https://github.com/cloudfoundry-incubator/fissile/blob/c617cf6372fe75ebe255f2920c1dff6d6d0059ba/kube/registry_credentials.go#L23
Deployment fails. See more in slack thread https://cloudfoundry.slack.com/archives/C8RU3BZ26/p1551369471021700

jandubois:
Fixes #487.

The template only checks for an empty username, so the same logic should apply to a private registry that doesn't require authentication.

### DIFF
--- a/kube/pod.go
+++ b/kube/pod.go
@@ -48,7 +48,9 @@ func NewPodTemplate(role *model.InstanceGroup, settings ExportSettings, grapher 
 		// This role requires a custom service account
 		spec.Add("serviceAccountName", role.Run.ServiceAccount, authModeRBAC(settings))
 	}
-
+	if settings.CreateHelmChart {
+		spec.Get("imagePullSecrets").Set(helm.Block(`if ne .Values.kube.registry.username ""`))
+	}
 	// BOSH can potentially have an infinite termination grace period; we don't
 	// really trust that, so we'll just go with ten minutes and hope it's enough
 	spec.Add("terminationGracePeriodSeconds", 600)
--- a/kube/registry_credentials.go
+++ b/kube/registry_credentials.go
@@ -24,8 +24,8 @@ func MakeRegistryCredentials(settings ExportSettings) (helm.Node, error) {
 
 		value = `{{ printf "{%q:{%q:%q,%q:%q,%q:%q}}" ` +
 			`.Values.kube.registry.hostname ` +
-			`"username" (default "" .Values.kube.registry.username) ` +
-			`"password" (default "" .Values.kube.registry.password) ` +
+			`"username" .Values.kube.registry.username ` +
+			`"password" .Values.kube.registry.password ` +
 			`"auth" (printf "%s:%s" .Values.kube.registry.username .Values.kube.registry.password | b64enc) ` +
 			`| b64enc }}`
 	}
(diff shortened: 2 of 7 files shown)


kube/registry_credentials.go
```diff
@@ -37,6 +37,9 @@ func MakeRegistryCredentials(settings ExportSettings) (helm.Node, error) {
 		SetAPIVersion("v1").
 		SetKind("Secret").
 		SetName("registry-credentials")
+	if settings.CreateHelmChart {
+		cb.AddModifier(helm.Block(`if ne .Values.kube.registry.username ""`))
```
mook-as, line 41: Can we assume that we can't have a password without a user? (This would be true of docker hub, but the point of the issue is that we are having problems on non-standard stacks…)

(Also, I guess this can be written as `if not .Values.kube.registry.username` too, but which is clearer is debatable.)

Also need to check if various CRI backends are happy with passwordless user names.

kube/pod.go
```diff
@@ -48,7 +48,9 @@ func NewPodTemplate(role *model.InstanceGroup, settings ExportSettings, grapher
 		// This role requires a custom service account
 		spec.Add("serviceAccountName", role.Run.ServiceAccount, authModeRBAC(settings))
 	}
-
+	if settings.CreateHelmChart {
+		spec.Get("imagePullSecrets").Set(helm.Block(`if ne .Values.kube.registry.username ""`))
+	}
```
mook-as, line 53: I'm regretting that we didn't just pass `settings.CreateHelmChart` to `helm.Encoder` instead; if we did that we can probably push a bunch of the random kube logic to one place…

mook-as (approved):
Approved pending trying to use this on various CRI backends.

kube/pod.go
jandubois: Yeah, me too! Because I forgot initially to wrap the `helm.Block()` calls in the `settings.CreateHelmChart` conditional and spent some time figuring out why the tests failed.

On the other hand, it forces you to think about what to do for the plain kube files. I really wish we would get rid of them and deal with the tests in a different way. Then the whole issue would go away.

kube/registry_credentials.go
jandubois: The actual problem we are dealing with is docker hub, so we are good for that one.

I don't think it makes sense to have a password without a username, but it would be easy enough to change the condition to

```
if or (ne .Values.kube.registry.username "") (ne .Values.kube.registry.password "")
```

I don't know why passwordless usernames would be a problem for us though. You can specify them using the template, but if the backend doesn't support them, then you wouldn't specify them anyways, so either way the templates work as long as the information provided is correct.

mook-as: Good point re: matching the backend, so I think we're good. (Though the people pushing the images (us) may not be the people running the Kube deployments… but if we only have one they're hosed regardless of what this bit does)

jandubois:
> Approved pending trying to use this on various CRI backends.

Are you going to try on "various CRI backends" or do you expect somebody else to do this?

I would be fine merging this as soon as we have confirmation that it indeed fixes #487.

mook-as:
Just going to wait for #487, yeah.

PR #495, opened 2019-04-23
scripts/readiness-probe: Be quieter

mook-as:
Reduce the amount of detail when the readiness probe isn't ready, as too much irrelevant information was being emitted as things changed state.   Error messages about things like monitrc missing tend to end up being misleading when troubleshooting a broke installation.

Also fixes the docker test that is for some reason failing on my machine (but passing on CI).

### DIFF
--- a/scripts/dockerfiles/readiness-probe.sh
+++ b/scripts/dockerfiles/readiness-probe.sh
@@ -23,9 +23,10 @@ update_readiness() {
     local svcacct=/var/run/secrets/kubernetes.io/serviceaccount
     curl --silent \
         --cacert "${svcacct}/ca.crt" \
-        -H "Authorization: bearer $(cat "${svcacct}/token")" \
-        -H 'Content-Type: application/merge-patch+json' \
-        -X 'PATCH' \
+        --header "Authorization: bearer $(cat "${svcacct}/token")" \
+        --header 'Content-Type: application/merge-patch+json' \
+        --output /dev/null \
+        --request 'PATCH' \
         "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/$(cat "${svcacct}/namespace")/pods/${HOSTNAME}" \
         --data '{
             "metadata": {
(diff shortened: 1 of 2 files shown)


scripts/dockerfiles/readiness-probe.sh
```diff
@@ -23,9 +23,10 @@ update_readiness() {
     local svcacct=/var/run/secrets/kubernetes.io/serviceaccount
     curl --silent \
         --cacert "${svcacct}/ca.crt" \
-        -H "Authorization: bearer $(cat "${svcacct}/token")" \
-        -H 'Content-Type: application/merge-patch+json' \
-        -X 'PATCH' \
+        --header "Authorization: bearer $(cat "${svcacct}/token")" \
+        --header 'Content-Type: application/merge-patch+json' \
+        --output /dev/null \
```
mook-as, line 28: I believe this is the cause of the random JSON we sometimes see — it's the patch output (because `--silent` doesn't remove the output).

scripts/dockerfiles/readiness-probe.sh
```diff
@@ -40,14 +41,17 @@ if test -n "${FISSILE_ACTIVE_PASSIVE_PROBE:-}" ; then
     trap update_readiness EXIT
 fi
 
+if [ ! -r /etc/monitrc ] ; then
+    echo "Waiting for monit to start"
```
mook-as, line 45: Silence the misleading error about `/etc/monitrc` not existing (that makes it seem like something has really gone wrong, when it's just not finished running configgin yet)

scripts/dockerfiles/readiness-probe.sh
```diff
@@ -40,14 +41,17 @@ if test -n "${FISSILE_ACTIVE_PASSIVE_PROBE:-}" ; then
...
+
 # Grab monit port
 monit_port=$(awk '/httpd port/ { print $4 }' /etc/monitrc)
 
 # Check that monit thinks everything is ready
 curl -s -u admin:"${MONIT_PASSWORD}" http://127.0.0.1:"${monit_port}"/_status | gawk '
-    BEGIN                                                     { status = 0 }
-    $1 == "status" && $2 != "running" && $2 != "accessible"   { print ; status = 1 }
-    END                                                       { exit status }
+    $1 == "status" && $2 != "running" && $2 != "accessible"   { print "Waiting for monit to be ready"; exit 1 }
```
mook-as, line 54: Use the same message regardless of which job is failing (so things can be collapsed)

scripts/dockerfiles/readiness-probe.sh
f0rmiga: `curl` outputs the command verbosity to standard error and the response to standard out.
Another way of doing this is:

`curl ... 2> /dev/null` as an alternative to `--silent`.
`curl ... 1> /dev/null` as an alternative to `--output=/dev/null`.

_Just thinking out loud_

scripts/dockerfiles/readiness-probe.sh
f0rmiga: Isn't it weird to exit from `gawk`?

f0rmiga (commented):
It looks good. I only have one question.

scripts/dockerfiles/readiness-probe.sh
mook-as: Yep, but I wanted to be sure I'm not missing _other_ messages.

scripts/dockerfiles/readiness-probe.sh
mook-as: Not really; it's exiting `gawk` early from `gawk`. It just happens (separately) to also exit the surrounding shell when `gawk` returns non-zero.

f0rmiga: I see!

f0rmiga (approved):

PR #496, opened 2019-04-30
Additional validation on instance group configs in the role manifest

mook-as:
This PR does a few things:

- Converts the `app/validation` tests to be driven by the YAML files, to make it easier to add more tests
- Add validation of instance-group-specific configuration templates (including variable use and sorting)
- Change the validation of global configuration templates to only allow things that are in actually used jobs (as opposed to any job in any referenced release)
- Move some tests from `model/resolver` to `app/validation`; this makes it possible to load not-quite-valid manifests. We should eventually move validation (but not normalization) out of the resolver.
- Make `ErrorList` actually implement `error`, so that we can print it out easier (without having convert).

Note that this will require changes to the UAA & SCF role manifests to conform to the new checks.


model/resolver/resolver.go
```diff
@@ -609,3 +602,21 @@ clusterRoleLoop:
 	}
 	return errors
 }
+
+// calculateConfigurationTemplates caculates the global configuration templates
```
andreas-kupries: nit: `caculates` vs `calculates`

andreas-kupries (changes requested):
Typo.

app/validation.go
```diff
@@ -48,28 +63,110 @@ func (f *Fissile) Validate() validation.ErrorList {
...
+			propertyDefaults)...)
+		allErrs = append(allErrs, checkForUndefinedVariables(
+			fmt.Sprintf("instance_groups[%s].configuration.templates", instanceGroup.Name),
+			instanceGroup.Configuration.Templates,
+			variableUsage,
+			false)...)
+
+		// Collect all property defaults across all instance groups for a global
+		// check later.
+		for propertyName, defaults := range propertyDefaults {
```
f0rmiga: Could we pull this out into a function that we can share with https://github.com/cloudfoundry-incubator/fissile/pull/496/files#diff-d869de81fbd78443594a015394c848c7R314?

app/validation_test.go
```diff
@@ -2,64 +2,96 @@ package app
...
-		`instance-groups[myrole].configuration.templates[properties.tor.bogus]: Forbidden: Role-manifest duplicates opinion, remove from manifest`,
-		// checkForUndefinedBOSHProperties light, manifest - For the bogus property used above for checkOverridden
-		`role-manifest 'tor.bogus': Not found: "In any BOSH release"`,
-		`light opinion 'tor.bogus': Not found: "In any BOSH release"`,
-		`properties.tor.hostname: Forbidden: Light opinion matches default of 'localhost'`,
-
-		// `XXX`, // Trigger a fail which shows the contents of `actual`. Also template for new assertions.
+	require.NoError(t, err)
+
+	roleManifestDir, err := os.Open(filepath.Join(workDir, "../test-assets/role-manifests/app/validation/"))
```
andreas-kupries, line 35: I had to go looking at the files in that directories to start understanding how you have made it data driven, with their sections essentially describing pieces or role manifest, opinions, and the errors expected from the validation code.

Maybe add a basic comment here to give an overview of things ?

app/validation.go
```diff
@@ -188,30 +308,22 @@ func checkForDuplicatesBetweenManifestAndLight(light map[string]string, roleMani
...
-				continue
+		for property, template := range instanceGroup.Configuration.Templates {
+			if template.IsGlobal {
+				if _, ok := check[property]; ok {
+					// Skip over duplicates of the global properties in the
+					// per-instance-group data, we already checked them in a
+					// different role
+					continue
+				}
+				check[property] = struct{}{}
```
f0rmiga: Every time I see `struct{}{}` I find it to be the ugliest thing, but at the same time, I love what it does.

andreas-kupries (commented):
Code may require a comment

model/resolver/resolver.go
mook-as: Fixed in 311c0c12dbf55e61bd2f9dc2332075f7a8afc131

app/validation_test.go
mook-as: Comment added (at the top of the function) in 311c0c12dbf55e61bd2f9dc2332075f7a8afc131; let me know if that's what you're looking for.

app/validation.go
mook-as: Fixed in 129e316b441ccd8d7603edb14ed8d4aeb0040238: We don't actually need this at all, we're just doing the same work twice. Thanks for spotting that.

andreas-kupries (approved):

app/validation.go
```diff
@@ -122,39 +203,62 @@ func checkParentsOfUndefined(p string, bosh propertyDefaults) bool {
...
-	// And the global properties
-	for _, propertyDef := range roleManifest.Configuration.Templates {
-		property := propertyDef.Key.(string)
-		template := propertyDef.Value.(string)
+	return allErrs
+}
+
+// checkForUndefinedVariables checks that all configuration templates are
+// defined in the variables section
+func checkForUndefinedVariables(label string, templates map[string]model.ConfigurationTemplate, variableUsage map[string]int, isGlobal bool) validation.ErrorList {
```
f0rmiga: For this one, we may want it as a method of a `validator` struct that initializes and hols the state of `variableUsage`. Also, move the `range templates` to the caller, as this function don't need the whole thing (law of Demeter, or principle of least knowledge). Then we eliminate the need for passing `isGlobal` as well.

app/validation.go
```diff
@@ -48,28 +63,94 @@ func (f *Fissile) Validate() validation.ErrorList {
...
+			}
+		}
+		previous = key
+	}
+
+	return allErrs
+}
+
+// checkForUndefinedProperties checks that all properties are known
+func checkForUndefinedProperties(label string, global bool, templates map[string]model.ConfigurationTemplate, knownProperties model.PropertyDefaults) validation.ErrorList {
```
f0rmiga: Here, we can make this function a method of a `validator` struct that takes the `knownProperties` as part of the initialization, and the `range templates` can happen on the caller side, so we don't need to pass the whole templates, neither the `global` parameter.

f0rmiga (changes requested):

app/validation.go
mook-as: Changed in 9b6136ede07c51872f28e995c24e15c78b4978df

app/validation.go
mook-as: Changed in 9b6136ede07c51872f28e995c24e15c78b4978df

app/validation.go
```diff
@@ -6,109 +6,198 @@ import (
...
 	}
 
-	boshPropertyDefaultsAndJobs := f.collectPropertyDefaults()
-	darkOpinions := model.FlattenOpinions(opinions.Dark, false)
-	lightOpinions := model.FlattenOpinions(opinions.Light, false)
-	manifestProperties := collectManifestProperties(f.Manifest)
+	return &validator{
+		errOut:           errOut,
+		f:                f,
+		propertyDefaults: f.collectPropertyDefaults(),
```
f0rmiga: I don't think this belongs to the validator. It's weird that we have it being passed to `checkForUndefinedProperty`, e.g.

We can call `f.collectPropertyDefaults()` and then pass the result to `checkForUndefinedProperty`.

And for `checkLightDefaults`, we pass it as a parameter.

app/validation.go
```diff
@@ -6,109 +6,198 @@ import (
...
 
-	// All properties must be defined in a BOSH release
-	allErrs = append(allErrs, checkForUndefinedBOSHProperties("role-manifest",
-		manifestProperties, boshPropertyDefaultsAndJobs)...)
+	for propertyName, templateDef := range v.f.Manifest.Configuration.Templates {
+		if templateDef.IsGlobal {
+			v.checkForUndefinedProperty(
+				"configuration.templates",
+				propertyName,
+				v.propertyDefaults)
```
f0rmiga: This is the weird part.

app/validation.go
```diff
@@ -6,109 +6,198 @@ import (
...
+	// some template, and all templates must come from a used job
 
 	// All dark opinions must be configured as templates
-	allErrs = append(allErrs, checkForUntemplatedDarkOpinions(darkOpinions,
-		manifestProperties)...)
+	v.checkForUntemplatedDarkOpinions()
 
 	// No dark opinions must have defaults in light opinions
-	allErrs = append(allErrs, checkForDarkInTheLight(darkOpinions, lightOpinions)...)
+	v.checkForDarkInTheLight()
```
f0rmiga, line 88: > And the light shineth in darkness; and the darkness comprehended it not.

app/validation.go
mook-as: This is because we call `v.checkForUndefinedProperty` elsewhere, with the instance-group-specific properties instead. I believe this is the only validation function we do this.

app/validation.go
mook-as: Fixed in a7c161c2d8fa267f03009f7b9f063996b53c8e4b (didn't realize we only use it outside `validate()` in one place).

app/validation.go
mook-as: Fixed in a7c161c2d8fa267f03009f7b9f063996b53c8e4b

f0rmiga (approved):

PR #497, opened 2019-05-02
Revert "Merge pull request #492"

jandubois:
Revert because changes to the volumeClaimTemplates in stateful sets are not allowed during updates.


andreas-kupries (approved):
Unit tests worked.

PR #503, opened 2019-05-30
CAP-106 Skip validation of templated service account names

jandubois:
This allows the user to "tunnel through" a template expression to select the service account at deployment time based on the enabled features, e.g.

```
service-account: '{{if .Values.enable.feature}}feature{{else}}default{{end}}'
```

Any service account name generated by the template expression must also be used by at least one role in the chart unconditionally because will will not emit definitions for service account that are not used by the chart at all (normally only the service accounts for the tests).

### DIFF
--- a/kube/rbac.go
+++ b/kube/rbac.go
@@ -36,9 +36,9 @@ func NewRBACAccount(accountName string, config *model.Configuration, settings Ex
 		return nil, nil
 	}
 
-	// If we want to modify the default account, there's no need to create it
-	// first -- it already exists
-	if accountName != "default" {
+	// If we want to modify the default account, there's no need to create it first -- it already exists.
+	// We also don't create templated account names; any possible name must already be defined elsewhere.
+	if accountName != "default" && !(strings.HasPrefix(accountName, "{{") && strings.HasSuffix(accountName, "}}")) {
 		var instanceGroupNames []string
 		for instanceGroupName := range account.UsedBy {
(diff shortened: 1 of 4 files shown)


viovanov (changes requested):
Is this necessary? Can we just have another "post-deployment-setup" for Eirini only?

If we have to implement this mechanism, is it possible to make it more generic?
i.e. bring the ability to shortcut to rendering for any value

jandubois:
> Is this necessary? Can we just have another "post-deployment-setup" for Eirini only?

It is not necessary, we can duplicate the post-deployment setup job, but Satadru and I considered the role duplication worse than adding this mechanism because duplicated configuration will inevitably get out of sync over time (see also CAP-106).

The right solution would be to have a separate post-deployment-setup role specifically for eirini and not mix it in with the general post-deployment. That way it can be guarded by a `if_feature: eirini` gate and have its own service account. That would be my preferred solution if you are willing to implement it.

> If we have to implement this mechanism, is it possible to make it more generic?
> i.e. bring the ability to shortcut to rendering for any value

The mechanism already "works" for any value, at least in principle. This PR just disables validation for the templated service account (and prevents it from being created). This additional logic is specific to the particular setting and cannot be generalized with reasonable effort.

jandubois:
> The right solution would be to have a separate post-deployment-setup role specifically for eirini and not mix it in with the general post-deployment.

This is being implemented by https://github.com/SUSE/scf/pull/2470

PR #505, opened 2019-06-20
Always generate a serviceAccountName, even for "default"

jandubois:
Otherwise upgrades can fail when switching from a non-default to a default service account.


mook-as (approved):
Huh, surprised we didn't have a test case for this…

PR #506, opened 2019-06-24
Remove the helm.Match() methods

jandubois:
They were added in 678b0ba, but are overly complex, given that the only usage was locating specific resource kinds in a list of nodes. The list version of `Match()` also did not actually work (which didn't
matter, as it wasn't used anywhere).

This commit provides a much simplified version of `matchNodeInlist()` and gets rid of all the `Match()` methods. This restores `helm/config.go` to 100% test coverage, which was the original motivation for this change. 😄

### DIFF
--- a/kube/kube_test.go
+++ b/kube/kube_test.go
@@ -266,11 +266,11 @@ func getBasicConfig() (map[string]interface{}, error) {
 	return converted.(map[string]interface{}), nil
 }
 
-// matchNodeInList iterates through a list of nodes and returns the first one
-// where node.Match(target) succeeds
-func matchNodeInList(list []helm.Node, target helm.Node) helm.Node {
+// findKind iterates through a list of resources and returns the first one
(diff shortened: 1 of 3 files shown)


kube/kube_test.go
```diff
@@ -268,9 +268,9 @@ func getBasicConfig() (map[string]interface{}, error) {
 
 // matchNodeInList iterates through a list of nodes and returns the first one
 // where node.Match(target) succeeds
```
mook-as: Please update the comment.

mook-as (changes requested):
Basically looks fine (I'm assuming CI will catch any issues), just need to update the doc comment.

kube/kube_test.go
jandubois: Good catch! Commit amended.

mook-as (approved):

PR #508, opened 2019-08-01
Use a better check to see if a template variable is nil

jandubois:
The `(quote nil)` value changed in tiller between 2.11.0 and 2.14.0 from `"<nil>"` to the empty string, while `(typeOf nil)` continues to render as `<nil>`.

[jsc#CAP-593]


mook-as (approved):
LGTM, do we have tests for this? There was no matching test change, but maybe it was covered by existing tests… can't seem to spot it though.

PR #509, opened 2019-08-02
Adding a new flag in fissile build

bikramnehra:
## Description

This PR is to introduce:

1. Printing the name of the image to be built.
2. `--dry-run` flag to skip image build step.

## Test plan

Trigger `fissile build` with `--dry-run` and check whether the build part was skipped, also confirm whether image name was displayed.

### DIFF
--- a/builder/release_image.go
+++ b/builder/release_image.go
@@ -30,10 +30,10 @@ import (
 type ReleasesImageBuilder struct {
 	CompilationCacheConfig string
 	CompilationDir         string
-	DisplayImageName       bool
 	DockerNetworkMode      string
 	DockerOrganization     string
 	DockerRegistry         string
+	DryRun                 bool
 	FissileVersion         string
 	Force                  bool
 	Grapher                util.ModelGrapher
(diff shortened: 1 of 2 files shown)


builder/release_image.go
```diff
@@ -40,6 +40,7 @@ type ReleasesImageBuilder struct {
 	NoBuild                bool
 	OutputDirectory        string
 	RepositoryPrefix       string
+	ShowImageName          bool
```
f0rmiga: `--display-image-name` sounds more natural.

builder/release_image.go
```diff
@@ -275,6 +276,12 @@ func (j releaseBuildJob) Run() {
 		if err != nil {
 			return err
 		}
+
+		if r.ShowImageName {
+			r.UI.Printf("Image Name: %s\n", color.YellowString(imageName))
+			return nil
```
f0rmiga, line 283: Instead of returning here, I'd add another flag `--no-run` and then return `nil` if it is set after this `condition`.

f0rmiga (changes requested):

f0rmiga:
It has been decided over a private channel that we want this functionality as a separate subcommand rather than a flag.

bikramnehra:
It is finally decided to introduce a `--dry-run` flag which will skip the build part. Also, name of the image to be built will always be printed.

f0rmiga (approved):

PR #511, opened 2019-08-28
Update tests to deal with url.Parse() changes from CVE-2019-14809

jandubois:


jandubois:
Added `in progress` label because the Go version on the CI pipelines has not yet been updated to one that includes the CVE fixes.

jandubois:
Implemented as part of #515

PR #515, opened 2019-09-21
kube: Bump resource versions to what kube 1.16 wants

### LINKED ISSUE #514: Generated K8s resources will not be supported in K8s 1.16
As announced [here](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/), deprecated APIs are being removed from K8s 1.16, which is scheduled to be released tomorrow, 18 Sep 2019 as specified [here](https://github.com/kubernetes/sig-release/tree/master/releases/release-1.16#tldr).

The APIs that affect fissiled helm charts are statefulsets and deployments. Fissile should generate them in `apps/v1` API and not `apps/v1beta1` or `extensions/v1beta1`

mook-as:
- `extensions`/`v1beta1/Deployment` moved to `apps/v1`
- `extensions/v1beta1`/`PodSecurityPolicy` moved to `policy/v1beta1`
- `apps/v1beta1`/`StatefulSet` moved to `apps/v1`
- Add support for fallback versions (to run on older kubes), even though we probably no longer support those obsolete kube clusters.

Fixes #514

### DIFF
--- a/kube/utils.go
+++ b/kube/utils.go
@@ -75,6 +75,23 @@ func (b *ConfigBuilder) SetAPIVersion(apiVersion string) *ConfigBuilder {
 	return b
 }
 
+// SetConditionalAPIVersion sets the kube API version of the resource to build;
+// if that API version is not available, use a fallback instead (to be
+// compatible with older releases of kube).  If we are not building a helm
+// chart, the desired API version is always used.
+func (b *ConfigBuilder) SetConditionalAPIVersion(apiVersion, fallbackAPIVersion string) *ConfigBuilder {
+	if b.settings == nil || !b.settings.CreateHelmChart {
+		b.apiVersion = apiVersion
+	} else {
+		b.apiVersion = fmt.Sprintf(
+			`{{ ternary "%s" "%s" (.Capabilities.APIVersions.Has "%s") }}`,
+			apiVersion,
+			fallbackAPIVersion,
+			apiVersion)
+	}
+	return b
+}
+
 // SetKind sets the kubernetes resource kind of the resource to build.
 func (b *ConfigBuilder) SetKind(kind string) *ConfigBuilder {
 	b.kind = kind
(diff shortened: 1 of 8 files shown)


kube/utils.go
```diff
@@ -75,6 +75,23 @@ func (b *ConfigBuilder) SetAPIVersion(apiVersion string) *ConfigBuilder {
...
+// SetConditionalAPIVersion sets the kube API version of the resource to build;
+// if that API version is not available, use a fallback instead (to be
+// compatible with older releases of kube).  If we are not building a helm
+// chart, the desired API version is always used.
+func (b *ConfigBuilder) SetConditionalAPIVersion(apiVersion, fallbackAPIVersion string) *ConfigBuilder {
+	if b.settings == nil || !b.settings.CreateHelmChart {
+		b.apiVersion = apiVersion
+	} else {
+		b.apiVersion = fmt.Sprintf(
+			`{{ ternary "%s" "%s" (.Capabilities.APIVersions.Has "%s") }}`,
```
jandubois: Do we document anywhere the minimum helm version required? I think `ternary` requires helm 2.10 (which seems sensible, even though we sometimes use 2.8.x).

Should we have a test in the generated charts to make sure the helm version is compatible?

mook-as: Looks like we're still testing against [2.8.2](https://github.com/SUSE/cf-ci/blob/develop/cf-ci-orchestration/Dockerfile#L41), hmm.

mook-as: Okay, so it turns out `helm template` does _not_ support `.Capabilities.APIVersions.Has` style checks, so we can't just run that to see if we have support for older clients.

Will need to get rid of `ternary`, I guess. (But yes, that doesn't work, though a good document on what versions we support, and figuring out how to test them, would be nice)

mook-as: Removed `ternary` in c642ece2c6e51fbd825c2e371a924a09ecdefcd4. Kind of annoyed that I can't actually test with `helm` in a test, though.

jandubois (approved):
Looks good to me (but I didn't actually test it on a kube 1.16 system)

PR #516, opened 2019-09-28
one liner to git clone with --recurse-submodules

drnic:


jandubois (approved):
Thanks you for your PR!

PR #522, opened 2019-10-14
Add KUBERNETES_CONTAINER_NAME env var

jandubois:
This is needed by configgin to export properties only from the container whose name matches the role name (and not the loggregator agent sidecar). There seems to be no mechanism to reference the `container.name` field in the env section, so the value has to be provided by fissile.

[jsc#CAP-761]

### DIFF
--- a/model/mustache.go
+++ b/model/mustache.go
@@ -78,6 +78,14 @@ func (g *InstanceGroup) GetVariablesForRole() (Variables, error) {
 		}
 	}
 
+	configs["KUBERNETES_CONTAINER_NAME"] = &VariableDefinition{
+		Name: "KUBERNETES_CONTAINER_NAME",
+		CVOptions: CVOptions{
+			Type:    CVTypeEnv,
+			Default: g.Name,
+		},
+	}
+
 	result := make(Variables, 0, len(configs))
 
 	for _, value := range configs {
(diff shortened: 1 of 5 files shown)


model/mustache.go
```diff
@@ -78,6 +78,14 @@ func (g *InstanceGroup) GetVariablesForRole() (Variables, error) {
 		}
 	}
 
+	configs["KUBERNETES_CONTAINER_NAME"] = &VariableDefinition{
+		Name: "KUBERNETES_CONTAINER_NAME",
+		CVOptions: CVOptions{
+			Type:    CVTypeEnv,
+			Default: g.Name,
```
mook-as, line 88: So the container name is the name of the instance group, not the container? Oh, because the "instance group" here does _not_ include the addons / colocated containers.

We should probably include a comment here so I don't get confused again in the future…

mook-as (changes requested):
Needs a comment.

Do we have any tests that would show what value we get for `KUBERNETES_CONTAINER_NAME` for colocated containers? We should probably get one of those… But maybe I just missed the relevant change.

model/mustache.go
jandubois: The co-located container has its own instance group; it is not part of the main instance group:

https://github.com/SUSE/scf/blob/develop/container-host-files/etc/scf/config/role-manifest.yml#L2590-L2591

What do you want the comment to say?

mook-as: Something to indicate that the `instance group` here _isn't_ the pod name because…

mook-as (approved):
:+1: waiting on CI before merge

PR #523, opened 2019-10-15
Return error when image is not present

alex-slynko:
errors.Wrap returns nil when error is not present.
Also suggest docker pull


jandubois (approved):

jandubois:
@alex-slynko Thank you!

PR #525, opened 2019-11-15
Pod startup ordering

jandubois:
# Implement configgin startup synchronization

The next version of configgin will export bosh properties to a secret and tag it with the value of `CONFIGGIN_VERSION_TAG` (the helm chart version plus the secrets rotation count).

Fissile now defines an environment variable that references that role's secret tag in each role that needs to import links from it. This has the effect of delaying startup of the pod until all links that have to be imported have been written to their corresponding secrets.

# Implement `ignore` option for link consumers

In the rolemanifest a job can now explicitly state that a specified bosh link will *not* be consumed:

```
      jobs:
      - name: sample-job
        release: sample
        consumes:
          not-used: {ignore: true}
```

The link definition will not be included in the job information passed on to Configgin, so the link will not resolve at runtime. It is the responsibility of the templates to not reference it (by setting properties, or by patching the templates).

Note that this feature allows ignoring of non-optional links as well!

### DIFF
--- a/model/instance_groups.go
+++ b/model/instance_groups.go
@@ -78,6 +78,11 @@ func (g *InstanceGroup) SetRoleManifest(m *RoleManifest) {
 	g.roleManifest = m
 }
 
+// Manifest return a reference to the instance groups role manifest
+func (g *InstanceGroup) Manifest() *RoleManifest {
+	return g.roleManifest
+}
+
 // CalculateRoleRun collects properties from the jobs run properties and puts them on the instance group
 // It also validates where necessary and is run *before* validateRoleRun
 func (g *InstanceGroup) CalculateRoleRun() validation.ErrorList {
--- a/model/instance_groups.go
+++ b/model/instance_groups.go
@@ -78,7 +78,7 @@ func (g *InstanceGroup) SetRoleManifest(m *RoleManifest) {
 	g.roleManifest = m
 }
 
-// Manifest return a reference to the instance groups role manifest
+// Manifest returns a reference to the instance groups role manifest
 func (g *InstanceGroup) Manifest() *RoleManifest {
 	return g.roleManifest
 }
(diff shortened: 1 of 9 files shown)


model/instance_groups.go
```diff
@@ -78,6 +78,11 @@ func (g *InstanceGroup) SetRoleManifest(m *RoleManifest) {
 	g.roleManifest = m
 }
 
+// Manifest return a reference to the instance groups role manifest
```
andreas-kupries: `return` --> `returns` (+ `s`).

test-assets/role-manifests/kube/volumes.yml
```diff
@@ -24,6 +24,17 @@ instance_groups:
           - path: /sys/fs/cgroup
             type: host
             tag: host-volume
+# The provider role only exist so that TestPodGetEnvVarsConfiggin can use
```
andreas-kupries: `exists`

test-assets/role-manifests/kube/volumes.yml
```diff
@@ -24,6 +24,17 @@ instance_groups:
...
+# it to reference in a mock up bosh link.
```
andreas-kupries: `to` -> `as` ? (Or remove `in` ?)

andreas-kupries (changes requested):
Just comment nits. Code LGTM.

model/instance_groups.go
jandubois: Addressed by fdbaeda

test-assets/role-manifests/kube/volumes.yml
jandubois: Addressed by fdbaeda

test-assets/role-manifests/kube/volumes.yml
jandubois: Addressed by fdbaeda

andreas-kupries (approved):
LGTM.

andreas-kupries:
Not merging yet, wait for `configgin` to be ready as well. Must be kept in close sync.

andreas-kupries (approved):
LGTM

PR #526, opened 2019-12-02
Define CONFIGGIN_SA_TOKEN for all pods using configgin

jandubois:
It points to the `configgin` service account token in a secret named `configgin` updated by a `configgin-helper` job. This variable is not defined for pods already running with a service account that is bound to the `configgin` role because it would make no difference, and more importantly would prevent bootstrapping of `configgin-helper`.

[jsc#CAP-828]

### DIFF
--- a/builder/role_image.go
+++ b/builder/role_image.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


andreas-kupries (approved):
LGTM

PR #527, opened 2019-12-05
Document cert subject names in values.yaml

jandubois:
In order to provide custom certs the user needs to know the subject names required for each cert. This commit adds them to the `values.yaml` file in the helm chart. For example:

```
  # PEM-encoded server certificate
  # This value uses a generated default.
  # This certificate uses the names "credhub-set" and
  # "server.dc1.{{.KUBERNETES_NAMESPACE}}.svc.{{.KUBERNETES_CLUSTER_DOMAIN}}".
  CREDHUB_SERVER_CERT: ~
```
Rules:

* The cert has neither role nor alternate names:
  This certificate uses the name "certificate-name".

* The cert has a role name but no alternate names:
  This certificate uses the role name "role".

* The cert has a role name and a single alternate name:
  This certificate uses the role name "role" and the additional name "alt".

* The cert has a role name and multiple alternate names:
  This certificate uses the role name "role" and the additional names "alt1", "alt2", and "alt3".

* The cert has no role but a single alternate name:
  This certificate uses the name "alt".

* The cert has no role but multiple alternate names:
  This certificate uses the names "alt1", "alt2", and "alt3".

The documentation will have to expand on how role names are mapped to a list of names, and which variables are used in template expansion.

### DIFF
--- a/kube/values.go
+++ b/kube/values.go
@@ -2,6 +2,7 @@ package kube
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"code.cloudfoundry.org/fissile/helm"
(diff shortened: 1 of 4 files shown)


andreas-kupries (approved):
LGTM

PR #528, opened 2019-12-07
Extend feature-flag support to colocated containers

tareqmamari:
There are situations where we need to enable/disable colocated
containers through feature flags, as currently it is not fully supported, and hence this PR.

### DIFF
--- a/kube/pod.go
+++ b/kube/pod.go
@@ -33,7 +33,9 @@ func NewPodTemplate(role *model.InstanceGroup, settings ExportSettings, grapher 
 			return nil, err
 		}
 
-		containers.Add(containerMapping)
+		node := helm.NewNode(containerMapping)
+		addFeatureCheck(candidate, node)
+		containers.Add(node)
 	}
 
 	imagePullSecrets := helm.NewMapping("name", "registry-credentials")


HeavyWombat (approved):
I tried to following setup:
- Use the `fissile` development setup from https://github.com/HeavyWombat/fissile-dev, which uses a `nginx-release` for the main container and a `loggregator-agen` release for the colocated container.
- Ran through the steps in the README to get a Helm Chart
- Deployed the Helm Chart with `kind` and checked the number of containers, which were as expected only set to 1.
- Re-deployed the Helm Charts with `--set enable.test_feature=true` and this ended up with two containers in the pod.

The key aspects I looked after were:
- In the `values.yaml`:
  ```
  enable:
    # The test_feature feature enables these instance groups:
    # loggregator_agent_colocated
    test_feature: false
  ```
- In the respective template file:
  ```
      {{- if .Values.enable.test_feature }}
      - env:
        - name: "CONFIGGIN_SA_TOKEN"
          valueFrom:
            secretKeyRef:
              name: "configgin"
              key: "token"
        [...]
        - name: "VCAP_SOFT_NPROC"
          value: {{ .Values.kube.limits.nproc.soft | quote }}
        image: "{{ .Values.kube.registry.hostname }}/{{ .Values.kube.organization }}/test-loggregator-agent-colocated:1a2c4e11fce90cedc9316c72a526c47a67a9fdb9"
        lifecycle:
          preStop:
            exec:
              command:
              - "/opt/fissile/pre-stop.sh"
        livenessProbe: ~
        name: "loggregator-agent-colocated"
        ports: ~
        [...]
      {{- end }}
  ```

It looks very nice.

tareqmamari:
Thanks @HeavyWombat for the review, @viovanov @mook-as @jandubois could you please have a look as well ?

jandubois:
@tareqmamari Yes, will take a look later today

jandubois:
@tareqmamari @HeavyWombat Please note that the current fissile `develop` requires an update to configgin-0.20. This has happened in the stemcells used by by `develop` in SCF, but if you use your own stemcells, you may need to bump configgin; otherwise deployment will hang.

PS: Just realized that I didn't push a `0.20.0` tag to the configgin repo when making the release to rubygems; will add the tag later today too.

jandubois (approved):

